PR #1, opened 2020-02-13
Add repo documentation

jazzdan:


README.md
```diff
@@ -1 +1,24 @@
...
+So you want to contribute an extension!
+
+We welcome additional extensions as pull requests.
+
+We want everyone to feel at home in this repo and its environs; please see our [Code of Conduct](CODE_OF_CONDUCT.md) for some rules that govern everyone's participation.
+
+## Extension Code Guidelines(?)
+Each extension should live in its own directory in the root of this repo and have at least a Tiltfile, and optionally other files. For a very simple example see the [hello world extension](https://github.com/windmilleng/tilt-extensions/tree/master/hello_world). For a more complex example, see the [Tilt Extension Authoring guide](TODO).
+
+Extension names much match these rules:
```
landism: This feels like a lot of low-level detail that isn't worth the reader's time, especially compared to the rest of the doc (like, as a reader, I start wondering about why this particular topic gets so much emphasis and if I'm missing something). It also feels like we're just putting the burden of some implementation detail's RFC on the user instead of simplifying it for them. Could we simplify these rules to something like "names must start with a letter, and can only contain alphanumeric characters and hyphens"? Or enforce this at review time, either manually or via CI? Or bury this in another doc that we link to from here? (for launch I'd be fine with just deleting this entirely and figuring something out later)

README.md
```diff
@@ -1 +1,24 @@
-# tilt-extensions
\ No newline at end of file
+# Tilt Extensions
```
landism: I suspect it'd make sense to make this page more about browsing / using extensions than authoring them. I wouldn't be surprised if page loads with intent to consume/browse extensions outnumber page loads with intent to contribute > 20:1 - maybe it'd make sense to put the contributing stuff in a section at the end, or in a separate doc?

It's probably not very important either way for when we're just showing it to people at the start, so I guess it doesn't need to block.

ellenkorbes: +1 prioritize browsing over contributing

README.md
jazzdan: I was planning on throwing together a quick CI, so I'll do that in a subsequent PR.

jazzdan: (and delete this in the meantime)

README.md
ellenkorbes: I suggest adding a link to some document that explains how to *use* an extension.

README.md
```diff
@@ -1 +1,18 @@
...
+This is the official Tilt Extension Repository. All extensions here have been vetted and approved by the Tilt Team.
+
+## Finding an Extension
+Here's a list of extensions along with a brief description. Try using your browsers search functionality to find an extension.
```
ellenkorbes: We should also group them by type later on.

README.md
```diff
@@ -1 +1,18 @@
...
+
+* **`hello_world`**: A barebones example. Prints "Hello world!" when `hi()` is called.
+
+## How to Contribute an Extension
+So you want to contribute an extension!
```
ellenkorbes: Unnecessary.

README.md
jazzdan: Agreed.

PR #4, opened 2020-02-21
extensions: procfile

landism:


procfile/Tiltfile
```diff
@@ -0,0 +1,20 @@
+def procfile_resources(procfile='Procfile', set_port=False, starting_port=5000):
+  """
+  Creates local resources from a Procfile (e.g., as defined by Foreman: http://blog.daviddollar.org/2011/05/06/introducing-foreman.html)
```
jazzdan: Should we link to a `local_resource` doc here?

procfile/Tiltfile
```diff
@@ -0,0 +1,20 @@
...
+
+  Args:
+    procfile (Str): the Procfile to load
```
jazzdan:
```suggestion
    procfile (Str): the path to the Procfile to load
```

PR #10, opened 2020-04-17
restart-container-alpha extension

maiamcc:
first pass at an extension for restart container (only works with
`docker_build` + k8s). I plan to leave it briefly undiscoverable and
ask a few user friends to try it.


maiamcc:
@nicks @landism ping

restart_container_alpha/README.md
```diff
@@ -0,0 +1,10 @@
+# Restart Container (Alpha)
```
nicks, line 1: do we want experimental extensions to have the name `alpha` in them? @landism ? feels odd to me

restart_container_alpha/Tiltfile
```diff
@@ -0,0 +1,63 @@
...
+                 "`restart_container()` (this extension is meant to REPLACE restart_container() )")
+
+    # rename the original image to make it a base image and declare a docker_build for it
+    base_ref = '{}{}'.format(ref, base_suffix)
+    docker_build(base_ref, context, **kwargs)
+
+    # declare a new docker build that adds a static binary of entr (previously
+    # compiled by tilt) to the user's image
+    df = '''
+    FROM tiltdev/entr as base
```
nicks, line 35: you probably want to pin a specific version of this extension to a specific tagged version of the `tiltdev/entr` image? If you do it this way, you'll run into intractable compatibility problems quick

maiamcc: good call, done!

restart_container_alpha/README.md
maiamcc: Hmm yeah i might just take off the alpha in the name and keep it unlisted until we've had a few ppl test it out.

maiamcc:
re: last commit: could have sworn `touch /.restart` was working fine before to trigger `entr`, but uh, maybe not? see [entr#32](https://github.com/eradman/entr/issues/32).

maiamcc:
For reference, some existing projects updated to use this extension:
- [tilt-example-go](https://github.com/windmilleng/tilt-example-go/tree/use-restart-extension)
- [servantes](https://github.com/windmilleng/servantes/tree/with-restart-ext)
- [abc123](https://github.com/windmilleng/abc123/tree/with-restart-ext)

Any other sample projects we have around that might be interesting to try this on?

maiamcc:
@nicks @landism made some changes/accounted for some edge cases, could i get an approval for landing this (un-advertised, going to test a bit more before I declare it DONE, write docs etc. -- but I want to be able to send it to people to test)

restart_process/Tiltfile
```diff
@@ -0,0 +1,76 @@
...
+        the child image / may conflict with args we specifically pass for the child image.
+    """
+    # since we'll be passing the `dockerfile_contents` when building the child image,
+    # remove any kwargs that might conflict
+    rm_key_if_exists(kwargs, 'dockerfile')
+    rm_key_if_exists(kwargs, 'dockerfile_contents')
+
+    # 'target' isn't relevant to our child build--it will just fail to find the
+    # specified stage and error
+    rm_key_if_exists(kwargs, 'target')
```
maiamcc, line 71: fwiw, noticed this interaction when dogfooding this extension with ClusterAPI. [Here's the ClusterAPI Tiltfile, using this extension](https://github.com/maiamcc/cluster-api/tree/restart-process-ext)

restart_process/Tiltfile
```diff
@@ -0,0 +1,76 @@
...
+    # compiled by tilt) to the user's image
+    df = '''
+    FROM tiltdev/entr:2020-16-04 as entr-img
+
+    FROM {}
+    RUN ["touch", "{}"]
+    COPY --from=entr-img /entr /
+  '''.format(base_ref, restart_file)
+
+    clean_kwargs(kwargs)
```
nicks, line 42: this is fine, but this might be cleaner & easier to follow if you copied the dict

```
dict([(k, v) for (k, v) in kwargs.items() if k not in blacklist])
```

it's a little bit weird/surprising to mutate kwargs in-place

restart_process/Tiltfile
```diff
@@ -0,0 +1,76 @@
...
+
+    # last live_update step should always be to modify $restart_file, which
+    # triggers entr to rerun $entrypoint
+    live_update.append(run('date > {}'.format(restart_file)))
+
+    docker_build(ref, context, entrypoint=entrypoint_with_entr, dockerfile_contents=df,
+                 live_update=live_update, **kwargs)
+
+
+def clean_kwargs(kwargs):
```
nicks, line 57: i had to do a double-take here that the `kwargs` param name here without the `**` operator, i thought that was frowned upon

restart_process/Tiltfile
```diff
@@ -0,0 +1,76 @@
...
+    clean_kwargs(kwargs)
+
+    # Change the entrypoint to use entr.
+    # entr allows you to run commands when files change: https://github.com/eradman/entr/
+    # this invocation says: whenever $restart_file changes, re-execute $entrypoint
+    entrypoint_with_entr = "echo '{}' | /entr -rz {}".format(restart_file, entrypoint)
+
+    # last live_update step should always be to modify $restart_file, which
+    # triggers entr to rerun $entrypoint
+    live_update.append(run('date > {}'.format(restart_file)))
```
nicks, line 51: i wouldn't expect this function to mutate its arguments. Probably better as:
live_update = live_update + ...

restart_process/Tiltfile
maiamcc: i would totally buy that!

PR #11, opened 2020-04-22
docs for restart_process extension

maiamcc:
Hello @landism,

Please review the following commits I made in branch maiamcc/restart-container-docs:

ec3e7d824e93bda035079f8805f8f23cb127258d (2020-04-22 13:04:58 -0400)
under the hood explanation

95b04b5bf9fed274426bcd4f64df93e324f230d5 (2020-04-22 12:54:18 -0400)
do footnotes work on github?

c9c4f605eaf0d3d1f23b8fa709e825399f3689e3 (2020-04-22 12:34:41 -0400)
usage docs

### DIFF
--- a/restart_process/README.md
+++ b/restart_process/README.md
@@ -63,9 +63,25 @@ run('date > /.restart-proc')
 Because `entr` will re-execute the entrypoint whenever `/.restart-proc'` changes, the above `run` step will cause the entrypoint to re-run.
 
 #### Provide `entr`
-For this all to work, the `entr` binary must be available on the Docker image.
+For this all to work, the `entr` binary must be available on the Docker image. The easiest solution would be to call e.g. `apt-get install entr` in the Dockerfile, but different base images will have different package mangers; rather than grapple with that, we've made a statically linked binary available on Docker image: [`tiltdev/entr`](https://hub.docker.com/repository/docker/tiltdev/entr).
 
+If you want to build `image-foo`, this extension will:
+- build your requested image as `image-foo-base` via `docker_build` call with all of your specified args/kwargs
+- build `image-foo` (the actual image that will be used in your resource) as a _child_ of `image-foo-base`:
+```Dockerfile
+FROM tiltdev/entr:2020-16-04 as entr-img
 
+FROM image-foo-base
+
+# we'll use this file to signal restarts,
+# make sure it exists
+RUN ["touch", "/.restart-proc"]
+
+# make the entr binary available on this image
+COPY --from=entr-img /entr /
+```
+
+Thus, the final image produced is tagged `image-foo` and has all the properties of your original `docker_build`, plus access to the `entr` binary.
 
 ## Unsupported Cases
 This extension does NOT support process restarts for:
--- a/restart_process/README.md
+++ b/restart_process/README.md
@@ -44,7 +44,28 @@ The call above looks just like the initial `docker_build` call except for one ad
 
 ## What's Happening Under the Hood
 
-TODO(maia) 👀
+This extension makes use of [`entr`](https://github.com/eradman/entr/), a utility for running arbitrary commands when files change. Specifically, we override the container's entrypoint with the following:
+
+```python
+echo '/.restart-proc' | entr -rz <entrypoint>
+```
+
+This invocation says:
+- when the container starts, run <entrypoint>
+- whenever the `/.restart-proc` file changes, re-execute <entrypoint>
+- `-z`: if <entrypoint> exits, `entr` should exit as well (so that k8s/Tilt can detect that the container has stopped)[¹](#fn1).
+
+We also set the following as the last `live_update` step:
+```python
+run('date > /.restart-proc')
+```
+
+Because `entr` will re-execute the entrypoint whenever `/.restart-proc'` changes, the above `run` step will cause the entrypoint to re-run.
+
+#### Provide `entr`
+For this all to work, the `entr` binary must be available on the Docker image.
+
+
 
 ## Unsupported Cases
 This extension does NOT support process restarts for:
@@ -57,3 +78,7 @@ Run into a bug? Need a use case that we don't yet support? Let us know---[open a
 
 ## For Maintainers: Releasing
 TODO(maia) 👀
+
+---
+### Notes
+<a name="fn1"></a>**1**: the `-z` flag is only available in `entr` >= [v4.5](https://github.com/eradman/entr/releases/tag/4.5)
--- a/restart_process/README.md
+++ b/restart_process/README.md
@@ -3,8 +3,57 @@
 🚨 Still Under Development! 🚨
 This extension is still actively in development, use at your own risk.
 
-This extension wraps a `docker_build` call such that at the end of a live_update, the container's process will rerun itself.
+This extension wraps a `docker_build` call such that at the end of a `live_update`, the container's process will rerun itself. (Use it in place of the `restart_container()` Live Update step, which doesn't work for all users and will soon be deprecated.)
 
-TODO: documentation
-TODO: release instructions
-TODO: custom_build support
+## When to Use
+Use this extension when you have an image (via `docker_build`) and you want to re-execute its entrypoint/command as part of a `live_update`.
+
+E.g. if your app is a static binary, you'll probably need to re-execute the binary for any changes you made to take effect.
+
+(If your app has hot reloading capabilities---i.e. it can detect and incorporate changes to its source code without needing to restart---you probably don't need this extension.)
+
+## How to Use
+
+Import this extension by putting the following at the top of your Tiltfile:
+```python
+load('ext://restart_process', 'docker_build_with_restart')
+```
+
+For the image that needs the process restart, replace your existing `docker_build` call:
+```python
+docker_build(
+    'foo-image',
+    './foo',
+    arg1=val1,
+    arg2=val2,
+    live_update=[x, y, z...]
+)
+```
+with a `docker_build_with_restart` call:
+```python
+docker_build_with_restart(
+    'foo-image',
+    './foo',
+    '/go/bin/foo',
+    arg1=val1,
+    arg2=val2,
+    live_update=[x, y, z...]
+)
+```
+The call above looks just like the initial `docker_build` call except for one added parameter, `entrypoint` (in this example, `/go/bin/foo`). This is the command that you want to run on container start and _re-run_ on Live Update.
+
+## What's Happening Under the Hood
+
+TODO(maia) 👀
+
+## Unsupported Cases
+This extension does NOT support process restarts for:
+- Docker images without shell support (e.g. `scratch`)
+- Images run in Docker Compose resources
+- `custom_build`
+- ???
+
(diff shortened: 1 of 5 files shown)


restart_process/README.md
```diff
@@ -1,10 +1,101 @@
...
+```python
+run('date > /.restart-proc')
+```
+
+Because `entr` will re-execute the entrypoint whenever `/.restart-proc'` changes, the above `run` step will cause the entrypoint to re-run.
+
+#### Provide `entr`
+For this all to work, the `entr` binary must be available on the Docker image. The easiest solution would be to call e.g. `apt-get install entr` in the Dockerfile, but different base images will have different package mangers; rather than grapple with that, we've made a statically linked binary available on Docker image: [`tiltdev/entr`](https://hub.docker.com/repository/docker/tiltdev/entr).
+
+If you want to build `image-foo`, this extension will:
```
landism, line 69:
```suggestion
To build `image-foo`, this extension will:
```

landism: ffti, but "if you want to" feels a bit more hypothetical or conditional or something

restart_process/README.md
```diff
@@ -1,10 +1,101 @@
...
+- build your requested image as `image-foo-base` via `docker_build` call with all of your specified args/kwargs
```
landism, line 70:
```suggestion
- build your requested image as `image-foo-base` via `docker_build`, with all of your specified args/kwargs
```

or add an "a" before `docker_build`

or I think this structure makes the important bit more clear:

- build the image via `docker_build` with all of your specified args/kwargs as normal, but with the name `image-foo-base`

restart_process/README.md
```diff
@@ -1,10 +1,101 @@
...
+- build `image-foo` (the actual image that will be used in your resource) as a _child_ of `image-foo-base`:
```
landism, line 71:
```suggestion
- build `image-foo` (the actual image that will be used in your resource) as a _child_ of `image-foo-base`:
```

Maybe add something like "but with an `entr` binary added"? I'd be tempted to even hide the dockerfile behind a "click to expand" thing.

restart_process/README.md
```diff
@@ -1,10 +1,101 @@
...
+
+# make the entr binary available on this image
+COPY --from=entr-img /entr /
+```
+
+Thus, the final image produced is tagged `image-foo` and has all the properties of your original `docker_build`, plus access to the `entr` binary.
+
+## Unsupported Cases
+This extension does NOT support process restarts for:
+- Docker images without shell support (e.g. `scratch`)
```
landism, line 89:
```suggestion
- Docker images without /bin/sh (e.g. `scratch`)
```

restart_process/README.md
```diff
@@ -1,10 +1,101 @@
...
+- Images run in Docker Compose resources
+- `custom_build`
+- ???
```
landism, line 92: Is this a TODO? Or if not, maybe we should add some context like "others? this is still alpha, let us know if you run into problems!" except that's kinda redundant with the next line, so maybe just remove this?

restart_process/README.md
```diff
@@ -1,10 +1,101 @@
...
+
+## For Maintainers: Releasing
+TODO(maia) 👀
 
-TODO: documentation
-TODO: release instructions
-TODO: custom_build support
+---
+### Notes
+<a name="fn1"></a>**1**: the `-z` flag is only available in `entr` >= [v4.5](https://github.com/eradman/entr/releases/tag/4.5)
```
landism, line 101: I'm not sure this will be very interesting to readers of the doc, since the doc doesn't describe any workflows in which users might be usign entr < 4.5? But w/e.

restart_process/README.md
```diff
@@ -1,10 +1,101 @@
...
+```python
+docker_build(
+    'foo-image',
+    './foo',
+    arg1=val1,
+    arg2=val2,
+    live_update=[x, y, z...]
+)
+```
+with a `docker_build_with_restart` call:
```
landism, line 33: it's probably worth linking to the function signature and/or using named arguments here, for clarity

restart_process/README.md
```diff
@@ -1,10 +1,101 @@
...
+    arg1=val1,
+    arg2=val2,
+    live_update=[x, y, z...]
+)
+```
+The call above looks just like the initial `docker_build` call except for one added parameter, `entrypoint` (in this example, `/go/bin/foo`). This is the command that you want to run on container start and _re-run_ on Live Update.
+
+## What's Happening Under the Hood
+
+This extension makes use of [`entr`](https://github.com/eradman/entr/), a utility for running arbitrary commands when files change. Specifically, we override the container's entrypoint with the following:
```
landism, line 48: I think it'd be good to have an intro along the lines of "The extension is designed to Just Work for you with just the knowledge above, so you can stop reading here. If you're curious about what the extension is actually doing, this section's for you!" (implied by "what's happening under the hood", but eh)

restart_process/README.md
```diff
@@ -1,10 +1,101 @@
...
+# make sure it exists
+RUN ["touch", "/.restart-proc"]
+
+# make the entr binary available on this image
+COPY --from=entr-img /entr /
+```
+
+Thus, the final image produced is tagged `image-foo` and has all the properties of your original `docker_build`, plus access to the `entr` binary.
+
+## Unsupported Cases
```
landism, line 87: maybe move this above the under the hood section, so people can stop reading when they hit that

restart_process/README.md
maiamcc: A+ much better ty

maiamcc:
Urk sorry to force-push, I rebased like a fool. @landism can you take another look?

restart_process/README.md
```diff
@@ -1,10 +1,101 @@
...
+echo "/.restart-proc" | entr -rz /bin/my-app
+```
+
+When specified as a `command` in Kubernetes or Docker Compose YAML (this is how Tilt overrides entrypoints), the above would therefore need to be executed as shell:
+```
+/bin/sh -c 'echo "/.restart-proc" | entr -rz /bin/my-app'
+```
+Any `args` specified in Kubernetes/Docker Compose are attached to the end of this call, and therefore in this case would apply TO THE `/bin/sh -c` CALL, rather than to the actual command run by `entr`; that is, any `args` specified by the user would be effectively ignored.
+
+In order to make `entr` executable as ARGV rather than as shell, we have wrapped it in a binary that can be called directly and takes care of the piping under the hood.
```
maiamcc, line 89: I'm not actually sure if this is the right language (in particular "executable as ARGV") -- what should this be?

restart_process/README.md
```diff
@@ -1,10 +1,123 @@
...
+
+We also set the following as the last `live_update` step:
+```python
+run('date > /.restart-proc')
+```
+
+Because `tilt-restart-wrapper` will re-execute the entrypoint whenever `/.restart-proc'` changes, the above `run` step will cause the entrypoint to re-run.
+
+#### Provide `tilt-restart-wrapper`
+For this all to work, the `entr` binary must be available on the Docker image. The easiest solution would be to call e.g. `apt-get install entr` in the Dockerfile, but different base images will have different package mangers; rather than grapple with that, we've made a statically linked binary available on Docker image: [`tiltdev/entr`](https://hub.docker.com/repository/docker/tiltdev/entr).
```
landism, line 94:
```suggestion
For this all to work, the `entr` binary must be available on the Docker image. The easiest solution would be to call e.g. `apt-get install entr` in the Dockerfile, but different base images will have different package managers; rather than grapple with that, we've made a statically linked binary available on Docker image: [`tiltdev/entr`](https://hub.docker.com/repository/docker/tiltdev/entr).
```

restart_process/README.md
```diff
@@ -1,10 +1,123 @@
...
+To build `image-foo`, this extension will:
+- build your image as normal (via `docker_build`, with all of your specified args/kwargs) but with the name `image-foo-base`
+- build `image-foo` (the actual image that will be used in your resource) as a _child_ of `image-foo-base`, with the `tilt-process-wrapper` and its dependencies available
+
+Thus, the final image produced is tagged `image-foo` and has all the properties of your original `docker_build`, plus access to the `tilt-restart-wrapper` binary.
+
+#### Why a Wrapper?
+Why bother with `tilt-restart-wrapper` rather than just calling `entr` directly?
+
+Because in its canonical invocation, `entr` requires that the file(s) to watch be piped via stdin, i.e. it is invoked like:
```
landism, line 105: also, not all images have a shell?

restart_process/README.md
```diff
@@ -1,10 +1,123 @@
...
+echo "/.restart-proc" | entr -rz /bin/my-app
+```
+
+When specified as a `command` in Kubernetes or Docker Compose YAML (this is how Tilt overrides entrypoints), the above would therefore need to be executed as shell:
+```
+/bin/sh -c 'echo "/.restart-proc" | entr -rz /bin/my-app'
+```
+Any `args` specified in Kubernetes/Docker Compose are attached to the end of this call, and therefore in this case would apply TO THE `/bin/sh -c` CALL, rather than to the actual command run by `entr`; that is, any `args` specified by the user would be effectively ignored.
+
+In order to make `entr` executable as ARGV rather than as shell, we have wrapped it in a binary that can be called directly and takes care of the piping under the hood.
```
landism, line 116:
```suggestion
In order to make `entr` useable without a shell, this extension uses a simple binary that invokes `entr` and writes to its stdin.
```

how's that sound?

maybe worth hyperlinking "simple binary" to the wrapper?

restart_process/README.md
```diff
@@ -1,10 +1,123 @@
...
+
+Note: ideally `entr` could accept files-to-watch via flag instead of stdin, but (for a number of good reasons) this feature isn't likely to be added any time soon (see [entr#33](https://github.com/eradman/entr/issues/33)).
 
-TODO: documentation
-TODO: release instructions
-TODO: custom_build support
+## For Maintainers: Releasing
+If you have push access to the `tiltdev` on DockerHub, you can release a new version of this extension like so:
```
landism, line 121:
```suggestion
If you have push access to the `tiltdev` repository on DockerHub, you can release a new version of the binaries used by this extension like so:
```

restart_process/README.md
```diff
@@ -1,10 +1,123 @@
...
+/tilt-restart-wrapper --watchfile='/.restart-proc' <entrypoint>
+```
+
+This invocation says:
+- when the container starts, run <entrypoint>
+- whenever the `/.restart-proc` file changes, re-execute <entrypoint>
+
+We also set the following as the last `live_update` step:
+```python
+run('date > /.restart-proc')
```
landism, line 88: huh, it doesn't have touch, but it has date? wild.

maiamcc: No, I didn't even test this part of the flow on `scratch` because it failed earlier in the live update (at the sync step). I use `date` instead of `touch` because entr doesn't respond to timestamp changes, only to file writes (see [entr#32](https://github.com/eradman/entr/issues/32)).

maiamcc: in retrospect i should maybe add a comment

restart_process/README.md
maiamcc: ha nope, I gave up on making this shell-proof when scratch didn’t support Live Update out of the box whoops

currently the one non-shell-proof bit is `run('date > /.restart-proc')` but i don't have ideas off the top of my head about how to fix, so leaving it for now

PR #12, opened 2020-04-22
restart_process: try start.sh instead of entr

maiamcc:
since `entr` must be executed as shell for piping, it means we can't
pass it as a list of args, which means that a user's k8s yaml args
will never propogate up to the command they're running.

I tried reimplementing this extension using re/start.sh to see
how hard it would be. It was pretty easy, though there remains a
similar stumbling block: how to let a user specify an entrypoint
in shell, if they want, or as argv (and therefore compatible with
`args` in k8s yaml). it's not perfect, but the `shell` param is
a stopgap.

this implementation solves the problem of the previous implementation:
it's now possible to pass an `entrypoint` such that, if `args` are present
in the yaml, they actually apply to the correct executable

this works fine with servantes and example-go, and is throwing weird errors
in cluster-api, I'll dig in tomorrow

### DIFF
--- a/restart_process/Tiltfile
+++ b/restart_process/Tiltfile
@@ -1,4 +1,3 @@
-RESTART_FILE = '/.restart-proc'
 TYPE_RESTART_CONTAINER_STEP = 'live_update_restart_container_step'
 
 KWARGS_BLACKLIST = [
@@ -13,10 +12,9 @@ KWARGS_BLACKLIST = [
 
 
 def docker_build_with_restart(ref, context, entrypoint, live_update,
-                              base_suffix='-base', restart_file=RESTART_FILE, **kwargs):
+                              base_suffix='-base', shell=False, **kwargs):
     """Wrap a docker_build call and its associated live_update steps so that the last step
     of any live update is to rerun the given entrypoint.
-
 
      Args:
       ref: name for this image (e.g. 'myproj/backend' or 'myregistry/myproj/backend'); as the parameter of the same name in docker_build
@@ -24,7 +22,6 @@ def docker_build_with_restart(ref, context, entrypoint, live_update,
       entrypoint: the command to be (re-)executed when the container starts or when a live_update is run
       live_update: set of steps for updating a running container; as the parameter of the same name in docker_build
       base_suffix: suffix for naming the base image, applied as {ref}{base_suffix}
-      restart_file: file that Tilt will `touch` during a live_update to signal the entrypoint to rerun
       **kwargs: will be passed to the underlying `docker_build` call
     """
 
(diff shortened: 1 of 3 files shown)


restart_process/Tiltfile
```diff
@@ -13,18 +12,16 @@ KWARGS_BLACKLIST = [
 
 
 def docker_build_with_restart(ref, context, entrypoint, live_update,
-                              base_suffix='-base', restart_file=RESTART_FILE, **kwargs):
+                              base_suffix='-base', shell=False, **kwargs):
```
landism, line 15: need add `shell` to the arg comments below

is there value in adding this parameter vs telling them to pass entrypoint as array in string? it feels inconsistent with our other APIs

maiamcc: yeahhh i go back and forth. it IS inconsistent with our other APIs, but the other APIs aren't exactly intuitive and we expect this extension to have a much higher surface area than those APIs we're trying to be consistent with.

you're probably right tho :-/

maiamcc:
@landism take another look please? big changes:
- added `shell_path` param for user to optionally specify where their shell lives if it won't be found by the shebang in the re/start scripts (which is currently `#!/bin/sh` but i'm gonna change it to use `/usr/bin/env`). fixes e.g. ClusterAPI's case: they use distroless:debug, their shell lives at `/busybox/sh` and they don't have a `/usr/bin/env`)
- just removed ability to wrap the entrypoint as shell, because re/start.sh doesn't seem to support it ([see bug report in slack](https://tilt-dev.slack.com/archives/C9SA451CJ/p1587680271166700) sigh).  we eventually want users to be able to run arbitrary shell in conjunction with process restarts but this is... not that day :-/ and I wanted to merge SOMETHING

maiamcc:
Trying a different tack with #13 -- the start.sh solution makes it really hard for the user to specify an entrypoint of arbitrary shell. e.g. `start.sh /bin/sh -c 'echo hello world; /go/bin/stuff'` spins up processes:
1. `start.sh /bin/sh -c 'echo hello world; /go/bin/stuff'`
2. `/bin/sh -c 'echo hello world; /go/bin/stuff'`
3. `/go/bin/stuff`

Calling `restart.sh` kills process 2 above but not process 3, i.e. we don't actually kill the binary (which makes things weird when we then try to relaunch the binary -- e.g. "port already allocated" etc.)

Hopefully #13 will allow us to address BOTH these use cases:
A. user can use this extension in conjunction with `args` specified in k8s yaml (which this PR was originally intended to address)
B. user can use this extension to invoke arbitrary shell as their entrypoint

PR #13, opened 2020-04-24
restart_process: use pre-compiled bin of entr that doesn't need to be run as shell

maiamcc:
problem: using `entr` as a solution for restarting processes meant we had to run `echo filename | entr...` ie had to run wrapped in `sh -c` which meant that args specified in k8s yaml would never take effect (because they were applied to the `sh` call rather than the user's executable)

solution: compilee a wrapper for `entr` so we can invoke it without piping

### DIFF
--- a/restart_process/Tiltfile
+++ b/restart_process/Tiltfile
@@ -40,29 +40,36 @@ def docker_build_with_restart(ref, context, entrypoint, live_update,
     base_ref = '{}{}'.format(ref, base_suffix)
     docker_build(base_ref, context, **kwargs)
 
-    # declare a new docker build that adds a static binary of entr (previously
-    # compiled by tilt) to the user's image
+    # declare a new docker build that adds a static binary of tilt-restart-wrapper
+    # (which makes use of entr to watch files and restart processes) to the user's image
     df = '''
-    FROM tiltdev/entr:2020-16-04 as entr-img
+    FROM tiltdev/entr:2020-04-24 as entr-img
 
     FROM {}
     RUN ["touch", "{}"]
     COPY --from=entr-img /entr /
+    COPY --from=entr-img /tilt-restart-wrapper /
   '''.format(base_ref, restart_file)
 
     # Clean kwargs for building the child image (which builds on user's specified
-    # image and copies in entr). In practice, this means removing kwargs that were
-    # relevant to building the user's specified image but are NOT relevant to building
-    # the child image / may conflict with args we specifically pass for the child image.
+    # image and copies in Tilt's restart wrapper). In practice, this means removing
+    # kwargs that were relevant to building the user's specified image but are NOT
+    # relevant to building the child image / may conflict with args we specifically
+    # pass for the child image.
     cleaned_kwargs = {k: v for k, v in kwargs.items() if k not in KWARGS_BLACKLIST}
 
-    # Change the entrypoint to use entr.
-    # entr allows you to run commands when files change: https://github.com/eradman/entr/
-    # this invocation says: whenever $restart_file changes, re-execute $entrypoint
-    entrypoint_with_entr = "echo '{}' | /entr -rz {}".format(restart_file, entrypoint)
+    # Change the entrypoint to use `tilt-restart-wrapper`.
+    # `tilt-restart-wrapper` makes use of entr (https://github.com/eradman/entr/) to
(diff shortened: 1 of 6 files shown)


restart_process/Tiltfile
```diff
@@ -40,29 +40,36 @@ def docker_build_with_restart(ref, context, entrypoint, live_update,
     base_ref = '{}{}'.format(ref, base_suffix)
     docker_build(base_ref, context, **kwargs)
 
-    # declare a new docker build that adds a static binary of entr (previously
-    # compiled by tilt) to the user's image
+    # declare a new docker build that adds a static binary of tilt-restart-wrapper
+    # (which makes use of entr to watch files and restart processes) to the user's image
     df = '''
-    FROM tiltdev/entr:2020-16-04 as entr-img
```
landism, line 46: lol, oops

restart_process/release.sh
```diff
@@ -2,10 +2,13 @@
 
 set -e
 
-TIMESTAMP=$(date +'%Y-%d-%m')
+TIMESTAMP=$(date +'%Y-%m-%d')
 IMAGE_NAME='tiltdev/entr'
 IMAGE_WITH_TAG=$IMAGE_NAME:$TIMESTAMP
 
+# build our binary
+env GOOS=linux GOARCH=amd64 go build -o tilt-restart-wrapper main.go
```
landism, line 10: I'd expect this to need to be a static binary as well. Have you tested it on `scratch`?

restart_process/main.go
```diff
@@ -0,0 +1,44 @@
...
+	flag.Parse()
+
+	args := os.Args[1:]
+	if *watchFile == "" {
+		*watchFile = defaultWatchFile
+	} else {
+		// user passed this arg, make sure we don't pass it along to entr
+		args = os.Args[2:]
+	}
+	fmt.Println("Hello, world.")
```
landism, line 26: 👀

restart_process/main.go
```diff
@@ -0,0 +1,44 @@
...
+	"syscall"
+)
+
+var defaultWatchFile = "/.restart-proc"
+var watchFile = flag.String("watchfile", "", "File that entr will watch for changes; changes to this file trigger entr to rerun the command(s) passed")
+
+func main() {
+	flag.Parse()
+
+	args := os.Args[1:]
```
landism, line 19: I think you can pass `"/.restart-proc"` directly, as second argument for the default value in `flag.String` above, use `flag.Args()` below, and delete these 7 lines of code.

restart_process/main.go
```diff
@@ -0,0 +1,44 @@
...
+	"os"
+	"os/exec"
+	"strings"
+	"syscall"
+)
+
+var defaultWatchFile = "/.restart-proc"
+var watchFile = flag.String("watchfile", "", "File that entr will watch for changes; changes to this file trigger entr to rerun the command(s) passed")
+
+func main() {
```
landism, line 16: I think it's worth writing down why this exists. It seems like it might be annoying to reconstruct that knowledge.

(or maybe this is still just a POC and you'll worry about that when this all moves past experimental)

restart_process/main.go
landism: (and the line below is probably also debug logging?)

restart_process/release.sh
maiamcc: i... thiiink it is static? see: https://www.reddit.com/r/golang/comments/53nl4b/why_all_go_binaries_are_staticallylinked/

Also, it works on `scratch`. (Live update does NOT work on `scratch` but that's a different kettle o fish 😅 )

restart_process/main.go
maiamcc: whooop.

restart_process/main.go
maiamcc: ha yeah i just found `flag.Args()`. good call.

restart_process/Tiltfile
maiamcc: You mean we're not 16 months into the pandemic?

restart_process/main.go
maiamcc: yyyeahhh fair

maiamcc: Writing this now but will submit in a separate PR

PR #15, opened 2020-04-25
Added an extension and docs for running Conftest in Tilt

garethr:
@victorwuky as mentioned in the docs.

### DIFF
--- a/conftest/Tiltfile
+++ b/conftest/Tiltfile
@@ -1,0 +1,54 @@
+# -*- mode: Python -*-
+
+CONFTEST_ARGS = [
+    'all_namespaces',
+    'combine',
+    'data',
+    'fail_on_warn',
+    'input',
+    'namespace',
+    'output',
+    'trace',
+    'update',
+    'policy',
+]
+
+def conftest(path, name='conftest', policy='policy', **kwargs):
+    """
+    Test configuration files against Open Policy Agent policies using Conftest
+
+    Args:
+      path: path to file to test
+      name: a name for the resource, defaults to conftest. Useful if you want multiple invocations
+      all_namespaces: use rules from all namespaces
+      combine: combine all files passed in into one structure
+      data: a list of paths to load additional data from
+      fail_on_warn: fail even if only warnings are found
+      input: input type if different from that autodetected from the file extension
+      namespace: the namespace to use for rules, defauls to main
+      output: the output format, can be stdout, json, tap or table
+      trace: whether or not to show the full policy trace, useful for debugging
+      update: download policies based on local configuration or from explicit repositories
+      policy: where to find the policy files, defaults to a directory called policy
+    """
+
+    # we want to support being able to pass arguments through to the local_resource function
+    # but first we need to split any kwargs explicitly intennded for conftest out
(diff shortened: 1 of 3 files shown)


victorwuky:
Thanks @garethr for the submission 🙂. 

@landism: Can you please review?

conftest/Tiltfile
```diff
@@ -0,0 +1,50 @@
...
+    local_resource_kwargs = {k: v for k, v in kwargs.items() if k not in CONFTEST_ARGS}
+
+    # we build a string of the conftest arguments
+    args = ""
+    for key,value in conftest_kwargs.items():
+        args += "--%s=%s " % (key, value)
+
+    local_resource(
+        name,
+        deps=[path, 'policy'],
```
landism, line 47: IIUC, this should be the `policy` arg, rather than the string literal `'policy'`

conftest/Tiltfile
```diff
@@ -0,0 +1,50 @@
...
+
+    # we want to support being able to pass arguments through to the local_resource function
+    # but first we need to split any kwargs explicitly intennded for conftest out
+    conftest_kwargs = {k.replace("_", "-"): v for k, v in kwargs.items() if k in CONFTEST_ARGS}
+    local_resource_kwargs = {k: v for k, v in kwargs.items() if k not in CONFTEST_ARGS}
+
+    # we build a string of the conftest arguments
+    args = ""
+    for key,value in conftest_kwargs.items():
+        args += "--%s=%s " % (key, value)
```
landism, line 43: I think this needs different handling for boolean flags, e.g., `all_namespaces`, `combine`, `fail_on_warn`

conftest/Tiltfile
```diff
@@ -0,0 +1,50 @@
...
+      name: a name for the resource, defaults to conftest. Useful if you want multiple invocations
+      all_namespaces: use rules from all namespaces
+      combine: combine all files passed in into one structure
+      data: a list of paths to load additional data from
+      fail_on_warn: fail even if only warnings are found
+      input: input type if different from that autodetected from the file extension
+      namespace: the namespace to use for rules, defauls to main
+      output: the output format, can be stdout, json, tap or table
+      trace: whether or not to show the full policy trace, useful for debugging
+      update: download policies based on local configuration or from exlicit repositories
```
landism, line 31:
```suggestion
      update: download policies based on local configuration or from explicit repositories
```

conftest/Tiltfile
garethr: The arguments work with values of True or False, so this works. But Tilt won't do any validation that the passed values are booleans. I've just pushed a change to do that validation, so if you pass non-boolean values you'll get a nice Tilt error rather than relying on the Conftest error.

conftest/Tiltfile
garethr: Thanks, fixed.

conftest/Tiltfile
garethr: Good suggestion. I've made this explicit in the Tilt extension vs relying on the underlying default.

garethr:
@landism thanks for the comments, should be addressed now.

conftest/Tiltfile
landism: oh, cool. thanks!

PR #18, opened 2020-05-27
Add an extension for manipulating namespaces. Fixes https://github.com/tilt-dev/tilt/issues/3204

nicks:
Hello @landism, @victorwuky,

Please review the following commits I made in branch nicks/ch3204:

5e84d9be3a0375e808bbafe9bfd7c0ae7dd6f9d7 (2020-05-27 10:14:24 -0400)
Add an extension for manipulating namespaces. Fixes https://github.com/tilt-dev/tilt/issues/3204

### DIFF
--- a/namespace/Tiltfile
+++ b/namespace/Tiltfile
@@ -1,0 +1,71 @@
+# -*- mode: Python -*-
+
+def namespace_yaml(name):
+  """Returns YAML for a namespace
+
+  Args:
+    name: The namespace name. Currently not validated.
+
+  Returns:
+    The namespace YAML as a blob
+  """
+
+  return blob("""apiVersion: v1
+kind: Namespace
+metadata:
+  name: %s
+""" % name)
+
+def namespace_create(name):
+  """Returns YAML for a namespace
+
+  Args:
+    name: The namespace name. Currently not validated.
+  """
+  k8s_yaml(namespace_yaml(name))
+
+def namespace_inject(x, ns):
+  """Takes K8s yaml, sets its namespace to `ns`, and returns it as a blob.
+
+  This modifies the yaml in two ways:
+  1. Sets .metadata.namespace to `ns`
+  2. Sets ..template.metadata.namespace to `ns`
+     This ensures the namespace in, e.g., Deployment Pod Template Specs is
+     set, but might have false positives if you have a CRD with some other
+     element named 'template'.
+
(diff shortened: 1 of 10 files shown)


namespace/Tiltfile
```diff
@@ -0,0 +1,71 @@
...
+     This ensures the namespace in, e.g., Deployment Pod Template Specs is
+     set, but might have false positives if you have a CRD with some other
+     element named 'template'.
+
+  Args:
+    x: K8s yaml. Either a filename (string) or the yaml itself (Blob)
+    ns: The namespace to set the K8s objects to.
+
+  Returns:
+    Blob containing the K8s object as yaml, with namespaces set to `ns`.
```
landism, line 42:
```suggestion
    Blob containing the K8s objects as yaml, with namespaces set to `ns`.
```

namespace/Tiltfile
```diff
@@ -0,0 +1,71 @@
...
+    objects = read_yaml_stream(x)
+  elif type(x) == 'blob':
+    objects = decode_yaml_stream(x)
+  else:
+    fail('only takes string or blob, got: %s' % type(x))
+
+  return encode_yaml_stream([f(o) for o in objects])
+
+def _set_k8s_yaml_namespace(o, ns):
+  o['metadata']['namespace'] = ns
```
landism, line 58: Ideally we'd only do this to namespace-scoped objects. That seems hard, so maybe just a caveat in the docs not to use this on cluster-scoped objects?

landism: or, hmm, I just tried creating a namespace with metadata.namespace set, and kubectl apply simply ignored the namespace, so that's probably fine? I couldn't find official documentation that k8s ignores namespaces on cluster-scoped objects, but some related discussion [here](https://github.com/kubernetes-sigs/kustomize/issues/552).

namespace/test/test.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+cd $(dirname $0)
+tilt ci
```
landism, line 4: 🙌

namespace/Tiltfile
nicks: done

namespace/Tiltfile
nicks: added a caveat

PR #19, opened 2020-06-02
Add `helm_remote` extension

KOGI:
@victorwuky
The currently recommended way of installing remote helm files is:
```python
local("helm upgrade --install <chart> <repo>")
```

But this has a few shortcomings when running `tilt down`:
1. the helm chart is not uninstalled
1. any namespace created with helm's `--create-namespace` flag is not uninstalled (even when using `tilt down --delete-namespaces`

Another option is:
```python
local("helm pull <repo>/<chart> --untar --destination helm_deps/<chart>")
k8s_yaml(helm("helm_deps/<chart>", namespace=myNamespace))
```
This has the advantage of solving bullet #1 above, but still has other shortcomings
1. namespace must already exist in order to install there (`helm()` has no support for `--create-namespace`)
1. even if namespace already exists, it still may not be cleaned up when running `tilt down --delete-namespaces`
1. if the chart defines any CRDs, they won't be installed as part of your call to `helm()`

This extension solves all of the above. (see comments in code for a few technical caveats)

### DIFF
--- a/helm_remote/README.md
+++ b/helm_remote/README.md
@@ -1,0 +1,42 @@
+# Helm Remote
+
+Author: [Bob Jackman](https://github.com/kogi)
+
+Install a remotely hosted Helm chart in a way that it will be properly uninstalled when running `tilt down`
+
+## Usage
+
+Because tilt doesn't have a way to dynamically ignore files, you'll need to add `.helm` to your project's `.tiltignore`
+file in order to prevent recursive re-processing of the main Tiltfile when the helm cache changes/updates.
+
+#### Install a Remote Chart
+
+```py
+load('ext://helm_remote', 'helm_remote')
+helm_remote('myChartName')
+```
+
+##### Additional Parameters
+
+```
+helm_remote(chart_name, repo_url='', repo_name='', namespace='', version='', username='', password='', values=[], set=[])
+```
+
+* `chart_name` ( str ) – the name of the chart to install  
+* `repo_name` ( str ) – the name of the repo within which to find the chart (assuming the repo is already added locally)
+<br> if omitted, defaults to the same value as `chart_name`
+* `repo_url` ( str ) – the URL of the repo within which to find the chart (equivalent to `helm repo add <repo_name> <repo_url>`)
+* `namespace` ( str ) – the namespace to deploy the chart to (equivalent to helm's `--namespace <namespace> --create-namespace` flags)
+* `version` ( str ) – the version of the chart to install. If omitted, defaults to latest version (equivalent to helm's `--version` flag)
+* `username` ( str ) – repository authentication username, if needed (equivalent to helm's `--username` flag)
+* `password` ( str ) – repository authentication password, if needed (equivalent to helm's `--password` flag)
+* `values` ( Union [ str , List [ str ]]) – Specify one or more values files (in addition to the values.yaml file in the chart). Equivalent to the helm's `--values` or `-f` flags
+* `set` ( Union [ str , List [ str ]]) – Directly specify one or more values (equivalent to helm's `--set` flag)
+
+#### Change the Cache Location
+
+By default `helm_remote` will store retrieved helm charts in the `.helm` directory at your project's root.
+This location can be customized by calling `helm_remote_set_cache_dir(newDirectory)`.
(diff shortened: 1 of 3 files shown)


drubin:
We currently also do some hacks as you have mentioned above.  This would be super helpful thanks for turning it into a plugin!

Would it be possible to also add an `README.md` to the folder to show some usages/info? 

Some nice examples 

* https://github.com/tilt-dev/tilt-extensions/tree/master/conftest 
* https://github.com/tilt-dev/tilt-extensions/tree/master/min_tilt_version

helm_remote/Tiltfile
```diff
@@ -0,0 +1,47 @@
...
+        namespace = 'default'
+
+    if repo_url != '':
+        local('helm repo add %s %s' % (repo_name, repo_url))  # updates if already added
+
+    if namespace != '':
+        # avoid a namespace not found error
+        namespace_create(namespace)  # do this early so it manages to register before we attempt to install into it
+
+    local('rm -rf helm_deps/%s' % repo_name)
```
drubin, line 31: This requires that `helm_deps` is git ignored right? Do you think we could download these into a temp directory location? or explicitly state these should be ignored. 

This changes the current working directory which feels a little odd.

We currently do 

```
local('if [ ! -d ./tmp/1.40.2/rabbitmq-ha ]; then helm fetch stable/rabbitmq-ha --untar --untardir ./tmp/1.40.2 --version 1.40.2; fi', quiet=QUIET)
rabbitMQYaml = helm('./tmp/1.40.2/rabbitmq-ha',
```
This allows us to cache and version lock the helm charts, but I have never loved that `./tmp` was `.gitignored` but it was part of our setup.

KOGI: That's a good thought about the .gitignore file. It definitely doesn't _require_ that directory to be ignored, but I think it's probably a good idea to create a readme file for this sub-folder which suggests ignoring it.

As far as where the temp dir is located/what it's called. I'm 100% open to suggestions. I could also add an additional function for configuring the temp/cache location? Even so, I'm open to input on what a sane default might be.

drubin: I agree with it being documented and has sane defaults. 

What about the versioning? Do you think your current method signature would support versioning? (this is something we would love to see and would love to switch to a proper plugin instead of our custom stuff)

KOGI: @drubin as I'm adding the ability to configure the temp/cache location, I'm seeing that your comment was maybe intended to refer to `.tiltignore` rather than `.gitignore`? In that case, yes, you're 100% correct that this dir would need to be ignored in order to prevent infinite recursive triggers to reprocess the tiltfile. So far I haven't been able to find a way around that, yet. Seems like tilt needs a native `unwatch_file()` function for situations like this, or to allow runtime ignoring of dynamic filenames.

KOGI: Added https://github.com/tilt-dev/tilt/issues/3404

KOGI:
@drubin Check out the [latest commit](https://github.com/tilt-dev/tilt-extensions/pull/19/commits/5f6fdd9a0fc0ece804b8933c3d154407aa5b9c21)

helm_remote/README.md
```diff
@@ -0,0 +1,42 @@
+# Helm Remote
```
KOGI, line 1: The rendered version of this file can be [viewed here](https://github.com/KOGI/tilt-extensions/blob/helm_remote/helm_remote/README.md)

victorwuky:
Thanks @KOGI for the submission. Thank you @drubin for the review so far! 

@jazzdan already assigned himself to do this PR and will review further on the Tilt side. Appreciate the collaboration so far!

KOGI:
@jazzdan the latest commit should fix this

helm_remote/Tiltfile
```diff
@@ -0,0 +1,88 @@
...
+helm_remote_cache_dir = './.helm'
+
+
+def helm_remote_set_cache_dir(directory):
+    helm_remote_cache_dir = directory
+
+
+# TODO: =====================================
+#   if it ever becomes possible for loaded files to also load their own extensions
+#   this method can be replaced by `load('ext://namespace', 'namespace_create')
```
KOGI, line 15: tilt-dev/tilt#3452

PR #24, opened 2020-06-07
local_output extension

cyucelen:
In the tilt.build [issue](https://github.com/tilt-dev/tilt.build/issues/179#issuecomment-639739992), @jazzdan mentioned that tilt extensions may have a functionality to get the output of a `local` command.

 I also use this frequently to put all of my scripting in my Tiltfile.

### DIFF
--- a/local_output/Tiltfile
+++ b/local_output/Tiltfile
@@ -1,0 +1,2 @@
+def local_output(cmd):
+  return str(local(cmd)).rstrip('\n')
--- a/local_output/Tiltfile
+++ b/local_output/Tiltfile
@@ -1,2 +1,2 @@
-def local_output(cmd):
-  return str(local(cmd)).rstrip('\n')
+def local_output(command):
+  return str(local(command)).rstrip('\n')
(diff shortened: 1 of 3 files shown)


local_output/Tiltfile
```diff
@@ -0,0 +1,2 @@
+def local_output(command):
+  return str(local(command)).rstrip('\n')
```
virta-jasonmay, line 2: can you include support for the `quiet` keyword?

drubin: What about just passing all the params to the local command? This way its semi future proof?

victorwuky:
Thanks @cyucelen for the submission. We are looking at this currently.

local_output/Tiltfile
virta-jasonmay: Last I looked at Starlark, it didn't allow slurping kwargs. But worth taking another look.

local_output/Tiltfile
```diff
@@ -0,0 +1,2 @@
+def local_output(command):
```
KOGI, line 1: Suggest also adding another method to return the most recent exit code. This would allow something like `dirExists = local_exitCode('test -d ./some/path') == 0`

KOGI: or possibly returning a tuple:
```
result, exitCode = local_output(command)
```

KOGI: or possibly returning a tuple:
```
result, exitCode = local_output(command)
```

PR #25, opened 2020-06-08
Add `git_resource` extension

KOGI:


nicks:
This is cool!

I'm a bit concerned about the high-level API - it's heavily parameterized, which is a sign that it doesn't compose well with other APIs.

What do you think of a more composable API like:

```
path = git_checkout(url)
deployment_from_dir(path)
```

or even `deployment_create` (like namespace_create in the namespace extension)

You can still have `git_resource`. It would just be syntactic sugar over the other public functions, and you could remove some of the customization parameters like yaml and build_callback

KOGI:
@nicks how is the number of (largely optional) parameters here any different than the large number of params defined for functions like `docker_build()` (16 params), `local_resource()` (10 params), or `k8s_resource()` (7 params)? I'm not disagreeing, just trying to clarify/understand.

In any case, I definitely like the idea of exposing things like `git_checkout()` and `deployment_from_dir()` etc while still maintaining the high-level `git_resource()` that glues them all together in a useful way.

nicks:
Oh boy, talking about how to break up `docker_build` is like my favorite topic!! 

The reason it started out as a single call is because it maps to a single docker API call. But in retrospect, I think it might make sense to separate out how to build the docker context (e.g., ignores) from general build settings that you would want to share among multiple docker builds (like the network). 

I also have a long-rime dream of an `image_build` primitive that lets you set a builder config for different builders that improves on custom_build, such that docker_build was implemented as `image_build` under-the-hood

but ANYWAY, ya, i think it still makes sense to have a "package it all together" function for the common case, just that exposing some of the helper functions might loosen the config burden a bit

esatterwhite:
Could this be expanded to enable ways to load tilt extensions from repos other than `tilt-extensions` ?

KOGI:
@esatterwhite that seems to me like probably something that would need to be done in tilt itself.

KOGI:
@nicks how's this look since the latest commit?

nicks:
thanks! I'll also add this to the tilt api ref.

PR #27, opened 2020-06-12
restart_process: readme tweaks and doc links

maiamcc:
Hello @landism,

Please review the following commits I made in branch maiamcc/restart-proc-readme:

1bad3466fa1afeff6bb71dbb38d78d7d9b3231c5 (2020-06-12 12:01:40 -0400)
restart_process: readme tweaks and doc links

### DIFF
--- a/restart_process/README.md
+++ b/restart_process/README.md
@@ -60,11 +60,15 @@ def docker_build_with_restart(ref: str, context: str,
 ```
 ## Unsupported Cases
 This extension does NOT support process restarts for:
-- Images run in Docker Compose resources
+- Images run in Docker Compose resources (use the [`restart_container()`](https://docs.tilt.dev/api.html#api.restart_container) builtin instead)
 - `custom_build`
 - Images without a shell (e.g. `scratch`, `distroless`)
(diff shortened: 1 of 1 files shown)


restart_process/README.md
```diff
@@ -60,11 +60,15 @@ def docker_build_with_restart(ref: str, context: str,
 ```
 ## Unsupported Cases
 This extension does NOT support process restarts for:
-- Images run in Docker Compose resources
+- Images run in Docker Compose resources (use the [`restart_container()`](https://docs.tilt.dev/api.html#api.restart_container) builtin instead)
 - `custom_build`
 - Images without a shell (e.g. `scratch`, `distroless`)
+- Container commands specified as `command` in Kubernetes YAML will be overridden by this extension.
+  - However, the `args` field is still available; [reach out](https://tilt.dev/contact) if you need help navigating the interplay between Tilt and these YAML values
```
maiamcc, line 67: this feels like it's worth mentioning otherwise ppl might be surprised, but i'm not sure this comment is more clarifying than confusing? thoughts?

landism: eh, I think it's fine

PR #35, opened 2020-06-13
feat: Allow pack to utilize it's own cache reuse optimizations

zetaron:
This expands on the `pack` extension to allow it to utilize `pack`s native cache reuse capabilities.

`pack` is only able to utilize caches if the docker image tag is the same, therefore this PR builds the `tilt-build-pack-caching` tag first and then retags the image to `$EXPECTED_REF`.

This PR helps my Team to reduce our day two `tilt up` times as initial builds are usually able to reuse previous caches.

some rudimentary measurements:
- after `docker system prune --all`: ~2 minutes
- after `docker system prune --all` + seeding pack related images: ~74 seconds
- using the pack cache mechanism: ~30 seconds


landism:
This is neat!

Unfortunately, IIUC, Tilt won't download the shell script when it gets the extension, so it won't work as-is.

The best workarounds I can think of, neither pretty, are to embed the shell script as a triple-quoted string in the Tiltfile and then either 1) write it out to a file on load or 2) use the whole string as the command for custom build.

If you go that path...you may already be familiar, but it took me a while to figure out, the best option I've found (also, still a bit messy) for writing a shell script in a Tiltfile is a quoted heredoc, e.g.:
```local('''cat <<'EOF' > build.sh
(script goes here)
EOF
```
so that the variables in the script get expanded at the appropriate time

zetaron:
@landism that hat is unfortunate, I did not know that Tilt *only* downloads the `Tiltfile`. I was assuming it would be the extensions whole directory. 

I will see to get around and refactor this into a heredoc and see if option 2 works out.

maiamcc:
FYI @zetaron I created an issue (https://github.com/tilt-dev/tilt/issues/3482) for extensions fetching the whole directory, not just the Tiltfile -- but I don't know if we'll be able to work on that any time soon, so in the meantime, you should definitely consider some workarounds (and let us know if we can help).

zetaron:
Hey @landism, @maiamcc a colleague of mine did some changes so that the extension is pure `Tiltfile` again with the added features. Maybe you could take another look?

PR #38, opened 2020-06-22
Avoid failure if there are no CRDs to install

KOGI:


KOGI:
@nicks @victorwuky

KOGI:
Just found out this actually breaks on mac... working on a workaround now...

nicks:
oof, i'll poke on this too. cross-platform bash is hard!

nicks:
Hmmm... @KOGI what problem was this PR meant to solve? I just double-checked the test I added, and it doesn't have any CRDs, and it seems to work ok?

KOGI:
@nicks what I was running into on my machine was that without the `-r` option, if no crd files were found, `xargs` was still running the `grep` command with 0 arguments. I don't recall anymore the exact details, but adding the `-r` option prevents calling `grep` if there are no results from the previous command.

I've got a fix for it (I think) that I'm just trying to get tested on a colleague's machine (I don't have a mac myself) -- hopefully will be posting a PR for it soon.

nicks:
@KOGI are you sure? when i tried running

```
$ echo "" | xargs grep -L xxx
(standard input)
```

i.e., grep with 0 arguments just pulled from standard input and used that (which i think this code already accounts for?)

nicks:
(though i'm not sure how portable grep is, or if there are distros where it will behave differently)

KOGI:
@nicks try this... do you see 2 output "echo" lines or just 1? (on my machine, I get 2):

```
echo -e "blah" | xargs -t echo > /dev/null && echo -e "" | xargs -t echo > /dev/null
```

nicks:
On my Ubuntu box, I get:

```
$ echo -e "blah" | xargs -t echo > /dev/null && echo -e "" | xargs -t echo > /dev/null
echo blah 
echo
```

On my macOS box, I get:

```
$ echo -e "blah" | xargs -t echo > /dev/null && echo -e "" | xargs -t echo > /dev/null
echo blah 
```

KOGI:
Right, so that 2nd line of output on your ubuntu box is what disappears by adding the `-r` option. I'll have to check out an old revision and spin up my cluster again and remember why that 2nd line was a problem to begin with.

KOGI:
@nicks ah, I remember now... if `grep` finds 0 matches, it returns an exit code of `1`, so if no yml files are found from the left half (which means `xargs` will call the 2nd `grep` with no arguments), then the 2nd `grep` call returns an error code, and tilt considers the command a failure. `-r` causes the 2nd call to `grep` to never be made when there are no arguments to pass.

KOGI:
@nicks see diff for new PR #46 for a proposed solution

PR #39, opened 2020-06-22
fix git checkout when specifying a branch name

KOGI:
Not sure how it ever worked before...

Before this change, if specifying a branch/tag name in your repo url as such: `git@github.com/owner/repo.git#branchName`, the command generated was akin to `git clone git@github.com/owner/repo.git#branchName -b branchName ./path/to/checkout`

Where the existence of that hash-tagged branch name was confusing to git (naturally so)

After this change, the same input results in a generated command of `git clone git@github.com/owner/repo.git -b branchName ./path/to/checkout`


KOGI:
@nicks @victorwuky

PR #40, opened 2020-06-22
determine the tilt root location before establishing file caches

KOGI:
Now that #3452 has been merged, it's possible for tiltfiles to be nested, and this breaks (in a way) both the `helm_remote` and `git_resource` extensions. With nested tiltfiles, it's now possible that these extensions create their private file caches in the nested location from which they're running, rather than in the tilt root directory. This PR fixes that.

### DIFF
--- a/git_resource/Tiltfile
+++ b/git_resource/Tiltfile
@@ -9,7 +9,8 @@ while not os.path.exists(os.path.abspath(path_prefix + 'tilt_modules')) and os.p
     path_prefix = '../' + path_prefix
 
 if os.path.exists(os.path.abspath(path_prefix + 'tilt_modules')):
-    git_resource_checkout_dir = os.path.abspath(path_prefix + git_resource_checkout_dir)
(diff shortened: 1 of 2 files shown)


KOGI:
@nicks @victorwuky

git_resource/Tiltfile
```diff
@@ -3,6 +3,15 @@
 # if customizing this location, you will also need to add the new location to your .tiltignore file to prevent infinite loops processing the main tiltfile (See github [issue #3404](https://github.com/tilt-dev/tilt/issues/3404))
 git_resource_checkout_dir = './.git-sources'
 
+# Find top-level Tilt path
+path_prefix = './'
+while not os.path.exists(os.path.abspath(path_prefix + 'tilt_modules')) and os.path.abspath(path_prefix) != '/':
+    path_prefix = '../' + path_prefix
```
nicks, line 9: os.path.dirname and os.path.join might work better here?

PR #42, opened 2020-06-24
restart-process: remove debug log line

jazzdan:


maiamcc:
😅  🙏  thanks @jazzdan !!

PR #46, opened 2020-06-25
helm_remote - fix grep command for mac

KOGI:
@nicks #38


KOGI:
@nicks I'm not a huge fan of resorting to platform detection for this, so I'm all ears if you've got better idea(s)

nicks:
hmmmm...doesn't the exit code problem also exist on mac though?

KOGI:
xargs on mac always runs with an implicit `-r` (which is why it's not an option on mac)

nicks:
ah ok. lgtm!

PR #47, opened 2020-06-30
support changing branches to one that may not exist in the local repo

KOGI:
if:
- a git resource has already been cloned/checkedout locally (pointing to "branch A")
- and a new branch (branch B) is pushed to origin
- the tiltfile is updated to use that branch

then:
- we get an error saying the new branch (branch B) was not found

----

This PR fixes that by fetching all from the remote server before attempting a checkout


KOGI:
@nicks @victorwuky

PR #50, opened 2020-07-06
feat: add release_name parameter to helm_remote

Tchoupinax:
Hello,

I added a new parameter for the function `helm_remote` in case of we need to name a release with a different name of its chart`s name.

My use case is that I have a chart called `base` and i launch two services `myServiceOne` and `myServiceTwo` from this same chart (the chart is the same but `values.yaml` are different.

In the current case, if i give `myServiceOne` as name the script will try to download `myRepo/myServiceOne` while the targeted chart is called `myRepo/base`. With this new option, targeted chart stays `base` but helm function is called with the wanted name. :)

Thank you for your work <3

### DIFF
--- a/helm_remote/README.md
+++ b/helm_remote/README.md
@@ -19,13 +19,15 @@ helm_remote('myChartName')
 ##### Additional Parameters
 
 ```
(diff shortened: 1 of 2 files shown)


jazzdan:
Thanks again @Tchoupinax!

PR #55, opened 2020-07-09
secret: add an extension for creating secrets

nicks:
Hello @jazzdan, @landism,

Please review the following commits I made in branch nicks/secret:

1f708350d271ae4ec2bbdda8030f098f7c81d441 (2020-07-09 14:58:00 -0400)
secret: add an extension for creating secrets

### DIFF
--- a/secret/Tiltfile
+++ b/secret/Tiltfile
@@ -1,0 +1,54 @@
+# -*- mode: Python -*-
+
+def secret_yaml_generic(name, namespace="", from_file=None):
+  """Returns YAML for a generic secret
+
+  Args:
+    name: The secret name.
+    namespace: The namespace.
+    from_file: Use the from-file secret generator. May be a string or a list of strings.
+       Example: ["ssh--privatekey=path/to/id_rsa", "ssh-publickey=path/to/id_rsa.pub"]
+
+  Returns:
+    The secret YAML as a blob
+  """
+
+  args = [
+    "kubectl",
+    "create",
+    "secret",
+    "generic",
+    name,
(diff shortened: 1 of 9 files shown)


nicks:
(i needed this to deploy some warehouse stuff, and it was easy enough to clean up and make an extension)

secret/Tiltfile
```diff
@@ -0,0 +1,43 @@
...
+      generator = " --from-file=%s" % from_file
+    elif type(from_file) == "list":
+      generator = " ".join(["--from-file=%s" % s for s in from_file])
+    else:
+      fail("Bad from_file argument: %s" % from_file)
+
+  if not generator:
+    fail("No secret generator specified")
+
+  return local("kubectl create secret generic %s %s %s -o yaml --dry-run=client" % (name, nsFlag, generator))
```
landism, line 32: I think we should pass this as an array of strings to avoid shell problems.

I don't think proper escaping is super important for most cases (like, can you even put a space in a k8s secret name?), but it's more important in extensions:
1. if someone gets a value from some other function and passes it as a name here and it happens to contain a space, without escaping the error message might be very confusing (like instead of "invalid name" maybe there's a bunch of gobbledygook)
2. It's good to set a good example of proper input handling
3. It's good to get comfortable with it ourselves so that we can ensure that people who do care about it can do it (like maybe we actually do want a shell-escape builtin). If we find it's kind of annoying we should fix that!

nicks: good point! i simply forgot we supported this. done!

PR #56, opened 2020-07-09
fix grep path for windows

KOGI:
@nicks @victorwuky @westongcooper

### DIFF
--- a/helm_remote/Tiltfile
+++ b/helm_remote/Tiltfile
@@ -83,6 +83,10 @@ def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], se
     # ======== Perform Installation
     local('rm -rf %s' % pull_target)
     local(pull_command)
+
+    if os.name == 'nt':  # windows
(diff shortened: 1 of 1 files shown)


nicks:
hang on, lemme test this on my windows machine

nicks:
hmmm...i think we need to do a lot more than this, when i tested this on windows
- xargs doesn't exist
- uname doesn't exist
- rm -fr doesn't exist

i think we might need to do something more clever here, gmas

KOGI:
@nicks you're right. I'm the original developer of this extension, but this entire time, I've neglected the fact that I'm actually running Tilt within MinGW which ends up being a (sometimes strange) mixture of linux/windows. In this particular case, Tilt's os.path* methods are detecting windows and converting paths to windows paths, but then those vars are being passed to local linux (mingw) commands. So `grep` resolves, but chokes on the windows paths.

Not really sure what to do here, then... if I add windows-equivalent commands, those will break inside mingw...

Suggestions?

PR #58, opened 2020-07-10
helm_remote: make all the tests pass on windows

nicks:
cc @KOGI 

what do you think of this as an alternative to https://github.com/tilt-dev/tilt-extensions/pull/56 ? does it work inside your mingw setup?

### DIFF
--- a/helm_remote/Tiltfile
+++ b/helm_remote/Tiltfile
@@ -63,13 +63,13 @@ def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], se
 
     # ======== Initialize
     # -------- targets
-    pull_target = '%s/%s' % (helm_remote_cache_dir, repo_name)
+    pull_target = os.path.join(helm_remote_cache_dir, repo_name)
     if version != '':
-        pull_target += '/%s' % version
+        pull_target = os.path.join(pull_target, version)
     else:
-        pull_target += '/latest'
+        pull_target = os.path.join(pull_target, 'latest')
 
-    chart_target = '%s/%s' % (pull_target, chart)
+    chart_target = os.path.join(pull_target, chart)
 
     # -------- commands
     pull_command = 'helm pull %s/%s --untar --destination %s' % (repo_name, chart, pull_target)
@@ -81,8 +81,10 @@ def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], se
         pull_command += ' --password %s' % password
 
     # ======== Perform Installation
-    local('rm -rf %s/%s' % (pull_target, chart))
+    local('rm -rf %s' % chart_target,
+          command_bat='if exist %s ( rd /s /q %s )' % (chart_target, chart_target))
     local(pull_command)
+
     install_crds(chart, chart_target)
 
     # TODO: since neither `k8s_yaml()` nor `helm()` accept resource_deps, sometimes the crds haven't yet finished installing before the below tries to run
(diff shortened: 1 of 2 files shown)


nicks:
(i ran the test script under msys/mingw and it seemed to work ok...)

helm_remote/Tiltfile
```diff
@@ -91,14 +93,11 @@ def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], se
     else:
         k8s_yaml(helm(chart_target, name=release_name, values=values, set=set))
 
-
 def install_crds(name, directory):
     name += '-crds'
+    files = str(local(r"grep --include '*.yaml' --include '*.yml' -rEil '\bkind\s*:\s+CustomResourceDefinition\s*$' %s || exit 0" % directory)).strip()
```
KOGI, line 98: does this not need a `command_bat` version?

nicks: there's a great grep implementation for windows! with good install packages for scoop, chocolatey, etc

KOGI:
@nicks I haven't had a chance to dig into it, yet, but this is what I'm getting on my machine:
```
  D:\code\lis-workspace\Tiltfile:17:8: in <toplevel>
  <builtin>: in include
  D:\code\lis-workspace\api-gateway\Tiltfile:3:12: in <toplevel>
  D:\code\lis-workspace\tilt_modules\helm_remote\Tiltfile:88:17: in helm_remote
  D:\code\lis-workspace\tilt_modules\helm_remote\Tiltfile:107:17: in install_crds
  <builtin>: in k8s_yaml
Error: json: line 0: invalid character '{' looking for beginning of object key string
```
And if we inspect the contents of the `files` variable, we have a list:
```
["D:\\code\\workspace\\.helm\\gloo\\1.3.28\\gloo/crds/auth_config.yaml", "D:\\code\\workspace\\.helm\\gloo\\1.3.28\\gloo/crds/gateway.yaml", "D:\\code\\workspace\\.helm\\gloo\\1.3.28\\gloo/crds/proxy.yaml", "D:\\code\\workspace\\.helm\\gloo\\1.3.28\\gloo/crds/route_table.yaml", "D:\\code\\workspace\\.helm\\gloo\\1.3.28\\gloo/crds/settings.yaml", "D:\\code\\workspace\\.helm\\gloo\\1.3.28\\gloo/crds/upstream.yaml", "D:\\code\\workspace\\.helm\\gloo\\1.3.28\\gloo/crds/upstream_group.yaml", "D:\\code\\workspace\\.helm\\gloo\\1.3.28\\gloo/crds/virtual_service.yaml", "D:\\code\\workspace\\.helm\\gloo\\1.3.28\\gloo/templates/100-crds.yaml"]
```

helm_remote/Tiltfile
```diff
@@ -91,14 +93,11 @@ def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], se
...
 
-    is_mac = str(local('uname -s')).strip().lower() == 'darwin'
-
-    files = str(local(r"grep --include '*.yaml' --include '*.yml' -rEil '\bkind\s*:\s+CustomResourceDefinition\s*$' %s" % directory + " | xargs" + ("" if is_mac else " -r") + " grep -L '^{{-'")).strip()
```
KOGI, line 100: The `grep -L '^{{-'` from this line was removed from the new line (98) above -- this is critical for filtering out any helm _template_ files that may contain the string `CustomResourceDefinition` but are simply templates rather than actual CRDs

KOGI:
@nicks see [this comment](https://github.com/tilt-dev/tilt-extensions/pull/58/files#r452983541)

helm_remote/Tiltfile
KOGI: If the goal is to obviate the call to `xargs`, this should work instead:
```
files = str(local(r"grep -L '^{{-' $(grep --include '*.yaml' --include '*.yml' -rEil '\bkind\s*:\s+CustomResourceDefinition\s*$' %s) || exit 0" % directory)).strip()
```

nicks:
@KOGI ah! good catch! I changed the test to use the gloo helm chart, which is more complex and tests this logic. Using `-L` in the grep command didn't work for me, but i guess we can just do this filtering in starlark

KOGI:
@nicks that seems to have done the trick!

nicks:
yay! thanks for checking

helm_remote/Tiltfile
KOGI: Sure, but that would mean the end-user would have to install it separately, right? Seems less than ideal (not that I know any other acceptable windows-friendly substitute...)

helm_remote/Tiltfile
```diff
@@ -91,18 +93,18 @@ def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], se
...
-    if (files == '') or (files == '(standard input)'):
+    if (files == ''):
         files = []
     else:
         files = files.split("\n")
 
+    # If the CRD files contain any Helm preprocessing, skip them.
+    files = [f for f in files if str(read_file(f)).find('{{-') == -1]
```
KOGI, line 106: ❤️

helm_remote/Tiltfile
nicks: yep!

In the medium-term, there are some Tilt APIs that should make this easier to do without `local()`. In the long-term, I'd love it if Tilt was smarter about ordering these CRDs, so that we didn't need to do this at all....

PR #59, opened 2020-07-10
helm_remote: remove specific chart instead of global directory

Tchoupinax:
Hello,

I noticed that at the end in my local folder `.helm/`, i have not all my remote chart, but only one. It is because the remove command remove the root folder instead of the specific chart. This PR fixes the issue :)

Thank you for your work !


jazzdan:
Thanks @Tchoupinax!

PR #62, opened 2020-07-14
restart_process: force root user so live update always works

### LINKED ISSUE #52: restart_process extension doesn't work if Docker image uses different user
If the target Docker image that the restart_process extension inherits `FROM` uses a user other than root then the resulting image will fail to build with the following error message:

```
Build Failed: ImageBuild: executor failed running [touch /.restart-proc]
```

Repro case: https://github.com/jazzdan/tilt-restart-process-permissions-error

(note how the Dockerfile here uses the `deno` user)

jazzdan:
Test forthcoming in the Tilt repo

Fixes #52

### DIFF
--- a/restart_process/Tiltfile
+++ b/restart_process/Tiltfile
@@ -11,7 +11,6 @@ KWARGS_BLACKLIST = [
     'target',
 ]
 
-
 def docker_build_with_restart(ref, context, entrypoint, live_update,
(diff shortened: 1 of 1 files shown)


jazzdan:
Test here https://github.com/tilt-dev/tilt/pull/3581

PR #63, opened 2020-07-15
tiltfile: added extension to enforce minimum k8s version

abdullahzameek:
Hello @nicks !

Here's the initial idea I came up with for enforcing a minimum tilt version. The functionality is heavily inspired by the min_tilt_version extension. Would something like this address the raised issue? (https://github.com/tilt-dev/tilt/issues/3297)

### DIFF
--- a/min_k8s_version/README.md
+++ b/min_k8s_version/README.md
@@ -1,0 +1,13 @@
+# Minimum Kubernetes Version
+
+Specify a minimum Kubenetes version to run this TiltFile
+
+If the minimum version is not met, Tilt will stop executing using the `fail` api
+
+## Usage
+
+
+```py
+load('ext://min_k8s_version', 'min_k8s_version')
+min_k8s_version('1.18.3')
+```
--- a/min_k8s_version/Tiltfile
+++ b/min_k8s_version/Tiltfile
@@ -1,0 +1,26 @@
+def min_k8s_version(min_version):
+
+   #This relies on the fact that the shell command returning something in the form 'gitVersion: vX.X.X'
+    version_str =  str(local('kubectl version --client  -o yaml| grep "gitVersion"', quiet=True))
+   
+   #Get the string, separate it at the colon, remove the 'v' and any spaces
+    version_str = version_str.split(':')[1].replace('v','').strip()
+  
+  #David Rubin's helper function returns the version in the form [X, X, X]
+    cur_ver = _version_tuple(version_str)
+    min_ver = _version_tuple(min_version)
+    
+    if cur_ver < min_ver:
+        print(" ---------------------------------------------------------- ")
+        print("|                                                          |")
+        print("|            Minimum Kubernetes Version Not Met!           |")
+        print("|                Refer link below to update                |")
+        print("| https://kubernetes.io/docs/tasks/tools/install-kubectl/  |")
+        print("|                                                          |")
+        print(" ---------------------------------------------------------- ")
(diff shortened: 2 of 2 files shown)


min_k8s_version/Tiltfile
```diff
@@ -0,0 +1,26 @@
+def min_k8s_version(min_version):
+
+   #This relies on the fact that the shell command returning something in the form 'gitVersion: vX.X.X'
+    version_str =  str(local('kubectl version --client  -o yaml| grep "gitVersion"', quiet=True))
```
nicks, line 4: i think you want the server version? --client means you're just getting the kubectl version

min_k8s_version/Tiltfile
```diff
@@ -0,0 +1,26 @@
...
+        print(" ---------------------------------------------------------- ")
+        print("|                                                          |")
+        print("|            Minimum Kubernetes Version Not Met!           |")
+        print("|                Refer link below to update                |")
+        print("| https://kubernetes.io/docs/tasks/tools/install-kubectl/  |")
+        print("|                                                          |")
+        print(" ---------------------------------------------------------- ")
+        fail('Minimum Kubernetes Version Not Met! Current Version is %s' % version_str)
+
+# From David Rubin's minimum tilt version extension
```
nicks, line 23: 👏

min_k8s_version/README.md
```diff
@@ -0,0 +1,13 @@
+# Minimum Kubernetes Version
```
nicks, line 1: can you also add this to the top-level README?

min_k8s_version/Tiltfile
abdullahzameek: Oh right, I'll fix that!

abdullahzameek: Fixed in 24cf22fbc413e681e6f928bf2a177c2ea23b4f1c !

min_k8s_version/README.md
abdullahzameek: Do you mean the main README here: https://github.com/tilt-dev/tilt-extensions/blob/master/README.md ?

nicks: yes

min_k8s_version/Tiltfile
```diff
@@ -0,0 +1,24 @@
...
+  #David Rubin's helper function returns the version in the form [X, X, X]
+    cur_ver = _version_tuple(version_str)
+    min_ver = _version_tuple(min_version)
+    
+    if cur_ver < min_ver:
+        print(" ---------------------------------------------------------- ")
+        print("|                                                          |")
+        print("|            Minimum Kubernetes Version Not Met!           |")
+        print("|                Refer link below to update                |")
+        print("| https://kubernetes.io/docs/tasks/tools/install-kubectl/  |")
```
nicks, line 17: this link isn't quite right. i think you're confusing the kubectl version (the local tool you use to access the cluster) with the kubernetes version (the particular server-side components)

in this case, the instructions to upgrade the cluster would depend a lot on what cluster you're using (docker for mac, etc).

It might be better to print out: cur_ver, min_ver, and k8s_context(), to catch the case where the user is connected to a different cluster than they expected

abdullahzameek:
Hey @nicks 

I've corrected the warning message in the following commit: 
f0aabce0a54e0549ebfa531ce3a1d59736ec886b  (Thu Jul 16 15:47:03 2020 +0400) -     corrected warning message

abdullahzameek:
Merged into master!

PR #64, opened 2020-07-16
restart_process: reduce image size

jazzdan:
Previous size (measured with `docker image inspect aaaaaa --format '{{.Size}}'`):
240996873 bytes = 240 MB

New size:
8546985 bytes = 8.5 MB

### DIFF
--- a/restart_process/Dockerfile
+++ b/restart_process/Dockerfile
@@ -7,8 +7,8 @@ RUN apk add build-base
 RUN git clone https://github.com/eradman/entr.git /entr
 RUN cd /entr; ./configure; CFLAGS="-static" make install
 
-FROM alpine:3.12
+FROM scratch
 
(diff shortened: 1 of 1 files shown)


jazzdan:
@maiamcc I _think_ I can. There was some trouble last time but I think it was resolved.

restart_process/Dockerfile
```diff
@@ -7,4 +7,8 @@ RUN apk add build-base
 RUN git clone https://github.com/eradman/entr.git /entr
 RUN cd /entr; ./configure; CFLAGS="-static" make install
 
+FROM alpine:3.12
```
landism, line 10: if you wanna get crazy, you could try making this `scratch`

jazzdan: Good idea! Done

PR #69, opened 2020-08-13
[secret] add ability to specify the type of secret

Tchoupinax:
Hello,

I added an option on secret extension to specify the type of secret. Indeed, this is necessary for `docker-registry` secret to be detected and correctly work. :)

It seems to be good with me. Any feedback is welcome !

Thank you for your work !

### DIFF
--- a/secret/Tiltfile
+++ b/secret/Tiltfile
@@ -1,6 +1,6 @@
 # -*- mode: Python -*-
 
-def secret_yaml_generic(name, namespace="", from_file=None):
+def secret_yaml_generic(name, namespace="", from_file=None, secret_type=None):
   """Returns YAML for a generic secret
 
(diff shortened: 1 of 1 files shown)


secret/Tiltfile
```diff
@@ -1,13 +1,15 @@
 # -*- mode: Python -*-
 
-def secret_yaml_generic(name, namespace="", from_file=None):
+def secret_yaml_generic(name, namespace="", from_file=None, secret_type=None):
```
nicks, line 3: i wish this could be `type=...`, but i see how that would conflict with the built-in operator!

Tchoupinax: Yeah i firstly used `type` of course but Python does not really love that. :x

PR #78, opened 2020-09-01
minor improvements to file_sync_only extension

nicks:
1) Make sure it works when Tilt is using a private registry (for KIND or Microk8s)
2) Make the tests part of the tilt-extensions CI

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ All extensions have been vetted and approved by the Tilt team.
 - [`api_server_logs`](/api_server_logs): Print API server logs. Example from [Contribute an Extension](https://docs.tilt.dev/contribute_extension.html).
 - [`conftest`](/conftest): Use [Conftest](https://www.conftest.dev/) to test your configuration files.
 - [`docker_build_sub`](/docker_build_sub): Specify extra Dockerfile directives in your Tiltfile beyond [`docker_build`](https://docs.tilt.dev/api.html#api.docker_build).
+- [`file_sync_only`](/file_sync_only): No-build, no-push, file sync-only development. Useful when you want to live-reload a single config file into an existing public image, like nginx.
 - [`git_resource`](/git_resource): Deploy a dockerfile from a remote repository -- or specify the path to a local checkout for local development.
 - [`hello_world`](/hello_world): Print "Hello world!". Used in [Extensions](https://docs.tilt.dev/extensions.html).
 - [`helm_remote`](/helm_remote): Install a remote Helm chart (in a way that gets properly uninstalled when running `tilt down`)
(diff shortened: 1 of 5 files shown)


nicks:
cc @MasayaAoyama made some minor tweaks, would you mind checking that it still works for you?

I'm very excited about this extension :grin:

MasayaAoyama:
I don't know how to recognize "using difference image registry" and $EXPECTED_REF env from python code.
But if I think we can, we will use the following code.

```diff
diff --git a/file_sync_only/Tiltfile b/file_sync_only/Tiltfile
index 080776f..9426227 100644
--- a/file_sync_only/Tiltfile
+++ b/file_sync_only/Tiltfile
@@ -14,15 +14,22 @@ def file_sync_only(image='', manifests=[], deps=["."], live_update=[]):
     # Check if tilt is using a different image registry.
     # If it is, we have to re-tag and re-push the image.
     build_script = """
-if [ "$EXPECTED_REF" != "%s*" ]; then
-  docker pull %s
-  docker tag %s $EXPECTED_REF
+if [[ ! "$EXPECTED_REF" =~ ^%s:* ]]; then
+  docker pull %s:%s
+  docker tag %s:%s $EXPECTED_REF
   docker push $EXPECTED_REF
 fi
-""" % (deployed_image, deployed_image, deployed_image)
+""" % (deployed_image, deployed_image, deployed_tag, deployed_image, deployed_tag)
+
+    if not "using difference image registry": # TODO
+        tag = deployed_tag
+    else:
+        # "part of $EXPECTED_REF's tag (or default value)" # TODO
+        tag = nil

     custom_build(deployed_image,
-                 build_script,
+        build_script,
+        tag=tag,
         deps=deps,
         skips_local_docker=True,
         live_update=live_update
@@ -44,7 +51,7 @@ def _get_current_tag(image='', manifests=[]):
             tags = tags + output.split("\n")

     if len(tags) == 0:
-        fail("cannot find image %s in manifests %s" % (image, manifests))
+        tags = ["latest"]
     elif len(tags) > 1:
         fail("found multiple image tags for %s in manifests %s" % (image, manifests))
```

MasayaAoyama:
Thank you for your review and merge my extension!  : )

nicks:
Hmmmm...yes, let me think about this a bit. I think there's some deeper issue. 

The way you're using custom_build here to NOT build the image is very interesting :). Our API for custom_build has a lot of parameters that interact in weird ways. This may just be a case where the interaction isn't quite right, and there needs to be a better way to handle it when you're using a private registry...

nicks:
ok! The latest version of Tilt (v0.17.4) has a fix that will make this work with a private registry, without any changes to the build script. But I'm going to merge this with @MasayaAoyama 's suggested change from above to set `tags = ["latest"]`, and with the tests

PR #79, opened 2020-09-02
Add extension for CoreOS kube-prometheus deployment

electrofelix:
To allow testing and development with services that make use of
PrometheusRule CRDs, provide an extension that is capable of deploying a
specific version with minimal patching in a manner that can
automatically pick up alerts defined in other namespaces.


electrofelix:
This is still fairly rough and could do with various improvements especially around the decision on where to cache files.

I'm thinking it might be worth checking if the user has a `${HOME}/.cache` directory available and defaulting to `~/.cache/tilt/downloads` or something like that and otherwise using `.tilt-cache` at the same level as the Tiltfile.

Then there are questions about use of a temp directory for the manifests, could add something like the following:
```
def _create_temp_dir():
    from_env = os.getenv('TILT_COREOS_PROMETHEUS_TEMP_DIR', '')

    if from_env != '':
        return from_env

    tmpdir = str(local("mktemp -d")).strip()
    os.putenv('TILT_COREOS_PROMETHEUS_TEMP_DIR', tmpdir)

    return tmpdir
```

That would help avoid polluting the cache with the extracted contents of the tarball.

coreos_prometheus/Tiltfile
```diff
@@ -0,0 +1,183 @@
...
+    return extract_path
+
+def patch_prometheus(path):
+    manifestsPath = os.path.join(path, 'manifests')
+    prometheusPatchesPath = os.path.join(os.path.dirname(__file__), 'prometheus')
+
+    # patches to modify certain behaviour
+    #  modify clusterrole for greater access
+    #  ensure all namespaces are scanned for alerts
+    localFiles = listdir(prometheusPatchesPath)
```
electrofelix, line 69: Could probably read in the yaml and modify the resulting objects instead of relying on files, depends on whether it's easier for someone to provide an additional path to check for patches, or if should expect objects defined in Tiltfile to be passed to manage this.

helm_remote/Tiltfile
```diff
@@ -98,6 +98,8 @@ def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], se
     else:
         k8s_yaml(yaml)
 
+    return yaml
```
electrofelix, line 101: I'm guessing the preference is to avoid modifying existing files, was trying to avoid needing to add a large blob of yaml to this repo at the same time.

This shouldn't break anyone?

.gitignore
```diff
@@ -1,2 +1,3 @@
 */test/.helm
-*/test/.git-sources
\ No newline at end of file
+*/test/.git-sources
+*/test/.tilt-cache
```
nicks, line 3: re: your bigger question about tilt caching -

ya, "fetch this remote tarball and cache the result" is turning out to be a pretty common pattern. 

I like your point about whether it should be a per-repo cache or a cache in $HOME/.cache.

I feel like most build systems I've worked with end up making some sort of a primitive for this to make this easy. But this seems fine for now.

nicks (commented):
nice!

I think the only question is how to handle the other files. (see my comment below). though i feel like we should just fix this in the tilt extension fetcher to fetch the whole directory.

helm_remote/Tiltfile
electrofelix: Cool

coreos_prometheus/README.md
```diff
@@ -0,0 +1,33 @@
+# CoreOS Prometheus
```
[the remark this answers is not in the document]
electrofelix: Done

coreos_prometheus/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+This extension deploys CoreOS Prometheus to a monitoring namespace with some
+modifications to the defaults to enable reading alerts defined via
+'PrometheusRule' custom resource definitions.
+
+## Usage
+
+Basic usage
```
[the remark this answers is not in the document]
electrofelix: I've refactored slightly so I think that I can avoid the reload issue.

coreos_prometheus/Tiltfile
```diff
@@ -0,0 +1,183 @@
...
+    # only copy file to final location to prevent redoing the above if the above
+    # have completed successfully.
+    local("mv %s.tmp %s" % (kube_prometheus_tarball, kube_prometheus_tarball))
+
+    # return path containing manifests directory
+    return extract_path
+
+def patch_prometheus(path):
+    manifestsPath = os.path.join(path, 'manifests')
+    prometheusPatchesPath = os.path.join(os.path.dirname(__file__), 'prometheus')
```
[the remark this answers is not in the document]
electrofelix: I haven't yet dealt with this issue, will wait a little longer to see if there is a better way than needing to patch inline.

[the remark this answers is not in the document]
nicks: ya, let me see if we can fix that quick

nicks:
the directory issue should be fixed in tilt v0.17.5! merging this

coreos_prometheus/Tiltfile
```diff
@@ -0,0 +1,192 @@
...
+    return results
+
+
+# To allow resources that require CRDs created by this extension to be selectively
+# pruned from other components if deploying of prometheus is disabled by the
+# file loading this one, need to always download and unpack the required files
+# and perform a limited amount of parsing.
+
+basedir = download_files()
+patch_prometheus(basedir)
```
electrofelix, line 177: discovered a bug here locally, on reload of the Tiltfile can end up in a constant loop. I'll need to test this code to be sure, but we've had to modify to ensure the files are only patched once.

I was previously thinking that if someone modified the files allowing to detect that they were modified and re-patching would be a good idea, but it doesn't quite work that way.

Below all the relevant files are read, and captured, so during a reload when the patch_prometheus is called to replace the files again, it ends up modifying files that are being watched resulting in a reload, which patches the files again. Either the patch needs to check if the files are different before replacing, or needs to be run once.

nicks: ya! a few extensions have this problem. the next version of Tilt is going to have a `watch_settings(ignore=...)` setting, so that any extension can tell tilt to ignore files. In the meantime, you can probably add a little note to the README like:

add `tilt_modules/coreos_prometheus/...` to your `.tiltignore`

PR #85, opened 2020-09-18
Initialize and update submodules during git checkout

alanchrt:
Hi. Some of our git repos use submodules. This change makes sure that submodules are always up to date (and as no effect on repos without submodules).

Something you guys would be open to merging upstream?


KOGI:
lgtm!

nicks (approved):

PR #87, opened 2020-09-18
Checkout fetched branch from origin on git update

alanchrt:
Another small pain point. At least with my git configuration, checking out a branch by name only checks out my local branch and not the one we just fetched from `origin`. So, it still needs a `git pull` to update after checkout or the code is still on my last-merged/fetched commit. This change just makes sure we always check out the `origin/*` ref for the branch that was just fetched.


nicks (approved):
lgtm!

PR #90, opened 2020-10-05
Add cert-manager extension

fabriziopandini:
Add an extension for deploying cert-manager and wait to it for being properly working by following the process described in the cert-manager documentation (https://cert-manager.io/docs/installation/kubernetes/#verifying-the-installation)

A working instance of cert-manager is required before calling `k8s_yaml` for any controller developed with Kubebuilder/using validation webhooks.

This extension makes it easier to get a working instance of cert-manager in the Tiltfile of all these projects

@victorwuky

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ This is the official Tilt Extensions Repository. Read more in [docs](https://doc
 All extensions have been vetted and approved by the Tilt team.
 
 - [`api_server_logs`](/api_server_logs): Print API server logs. Example from [Contribute an Extension](https://docs.tilt.dev/contribute_extension.html).
+- [`cert-manager`](/cert-manager): Deploys cert-manager.
 - [`conftest`](/conftest): Use [Conftest](https://www.conftest.dev/) to test your configuration files.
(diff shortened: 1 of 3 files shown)


victorwuky:
@nicks : Could you please review this? Thanks.

nicks:
Hi @fabriziopandini ! Thanks for the PR !!

I've been thinking about this.  I thought we heard some grumblings from other people on ClusterAPI-team about installing the cert manager synchronously with local() calls. Doing it this way means that Tilt can't parallelize and deploy other things while it's waiting on cert-manager to come up

There's been some backchannel discussion on how to make it faster with k8s_resource calls and local_resource calls
https://github.com/tilt-dev/tilt/issues/3710#issuecomment-692919161

alternatively, maybe it makes sense long-term for the extension to offer both a synchronous version and an asynchronous/parallelizable version, so it's ok if this is synchronous in the short-term

@maiamcc, what do you think? i know you had a bunch of opinions on this

fabriziopandini:
Hi @nicks!
I personally don't see the need for parallelism as paramount, because you are required to wait only on the first tilt-up.
however, no objection at all if someone wants to contribute an async mode

nicks:
sounds good to me! ya, we can add an async mode later

nicks (approved):

Issue #92, opened 2020-10-13
Exit code of restart_process is wrong 


jani123:
# Introduction

Restart process wrapper returns exit code zero even if software exits with any other. That affects on Tilt UI side so that service seems ok(green) even though it is not. Problem lies on used entr cli command, that seem to be returning its own exit status, instead of the status code from software it runs.

# Example

Lets have a script that fails on star
```
$ cat fail.sh 
#!/usr/bin/env bash

exit 1
```
And then we run restart process wrapper
```
$ go run tilt-restart-wrapper.go -entr_path=/usr/bin/entr -watch_file=test.go -- ./fail.sh; echo $?
hi there
0
```
The last line 0 is exit status of restart process wrapper, and in this case it should be 1 instead of 0.

This can be tracked to be fault(or feature depending on the view) of entr cli tool:
```
$ echo "./test.go" | entr -rzns ./fail.sh; echo $?
0
```
Here we have same failing script but the entr is only returns zeroes.

# How to almost fix this

One can use Kubernetes liveness properties in case there is multiple component starting up and they depend on eachother, but the most annoying part of this is that Tilt user interface lies about everything being ok even though it is not. We have 15 to 20 components in our software that is started up/developed and if one of them fails, you don't get any indication about it.

# Likely fix

My bet would be to get rid of the entr tool, since this is likely feature of the tool. It shouldn't be that hard to make golang code to [run application](https://golang.org/pkg/os/exec/) and restart it when [some files changes](https://github.com/fsnotify/fsnotify). What do you think? Unfortunately this is way too big of the job for now for me to do pull request since I have schedule of my own.

nicks:
cc @maiamcc do you remember the history behind this? i thought we had baked in a special version of entr that returned the right exit code, but i'm probably misremembering

maiamcc:
Oh ugh :-/ good catch @jani123. We don't have a special version of entr for the exit code-- @nicks you might be remembering that we baked in a version of entr with the `-z` flag (which wasn't yet in a release at the time the extension was published).

I (foolishly) didn't verify the exit code behavior because for a Pod, _any_ process exit is enough to alert Tilt that something weird is going on. Are you running into this bad behavior on Jobs/other shortlived k8s objects? Or on pods? Can you tell me a little more about the behavior you're seeing?

jani123:
[...]

PR #93, opened 2020-10-15
restart_process: use our fork of entr to return exit code of user command

maiamcc:
Addresses https://github.com/tilt-dev/tilt-extensions/issues/92

([Our fork of entr is here](https://github.com/tilt-dev/entr))

Test case: https://github.com/tilt-dev/failing-job


nicks:
would it make sense to put the test in this repo?

you can do something like
tilt ci || true > out.txt
and make sure the output about a failing job showed up in the log

nicks (approved):

PR #94, opened 2020-10-20
snyk: add additional testing capabilities

JimCodified:
This expands upon the original `snyk` extension by adding the following capabilities:

* Added the ability to run container and open source tests, in addition to iac tests
* Ability to supply additional options to the Snyk CLI
* Tests are set to TRIGGER_MODE_MANUAL so as to not fire off tests for every code change

@victorwuky
@shaimendel

### DIFF
--- a/snyk/Tiltfile
+++ b/snyk/Tiltfile
@@ -1,13 +1,13 @@
 
-def snyk(name, path, test_type, k8s_resource, *options):
+def snyk(path, test_type, name='snyk', k8s_dep='', *options):
     """
     Args:
-      name: a name for the resource. useful for labeling when running multiple tests.
-      path: path to the file or container image name:tag to test
+      path: path to the file or container image (name:tag) to test
       test_type: one of 'oss', 'container', or 'iac' corresponding to the various Snyk CLI tests
-      k8s_resource: used for container tests - sets the re-test dependency. ignored by other test types.
-      *options: additional CLI options; appended to the snyk command. see `snyk --help` for usage.
-          - container: this is where to insert the 'file==path/to/Dockerfile' option
+      name: a name for the resource. useful for labeling when running multiple tests. defaults to 'snyk'
+      k8s_dep: used for container tests - sets the re-test dependency for Tilt. Ignored by other test types.
+      *options: additional CLI options; appended to the snyk command. see `snyk --help` for full CLI options.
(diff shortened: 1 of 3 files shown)


JimCodified (commented):
Updated the general Tilt Extensions README to include updated info about `snyk` extension capabilities.

victorwuky:
@nicks can you please help review? Thanks!

snyk/Tiltfile
```diff
@@ -1,11 +1,41 @@
...
+        ])
+        test_deps = [k8s_dep]
+    elif test_type == 'iac':
+        snyk_cmd = " ".join([
+            "snyk iac test",
+            path
+        ])
+        test_deps = [path]
+    else:
+        print("Test type should be one of oss, container, or iac.")
```
nicks, line 31: should this be a `fail` command? i don't think we want this to fall through

snyk/Tiltfile
```diff
@@ -1,11 +1,41 @@
...
+
+    if (len(options) > 0):
+        snyk_cmd = " ".join([snyk_cmd, options[0]])
```
nicks, line 34: this only adds the first option? is that what you intended?

snyk/Tiltfile
```diff
@@ -1,11 +1,41 @@
...
+
     local_resource(
         name,
-        deps=[path],
-        cmd= 'snyk iac test %s' % (path)
+        deps = test_deps,
+        cmd = snyk_cmd,
```
nicks, line 39: now that this command is becoming more complex, i wonder if it makes sense to use an args array here instead of a string, so that a bad command doesn't cause bash parsing issues. it's ok to leave this as a TODO in this PR

nicks (commented):
Thanks! I'm ok with this if @shaimendel doesn't have any objections. i have a few minor Q's about the implementation though

snyk/Tiltfile
JimCodified: It's where my intentions ended up. Had a different idea in mind originally. Changed it to just a simple string.

snyk/Tiltfile
JimCodified: Fixed.

nicks (approved):

nicks:
thanks!

PR #99, opened 2020-11-03
Add a Ko extension

nicks:


ImJasonH:
Thanks for putting this together! I'll take some time to play with it this week.

I noticed something about the example that I wanted to get your thoughts on. Ko requires users to write their YAML such that there's a reference to the importpath wherever the built image should be replaced. So the YAML would say `image: ko://github.com/foo/bar/cmd/baz` (or an arg even, `[-foo, ko://github.com/foo/bar/cmd/baz]`), and `ko` searches the YAML for `ko://`-prefixed valid importpaths and replaces them.

So in your example, `ko_build('hello-world-image', 'github.com/foo/bar/cmd/baz')` should be `ko_build('ko://github.com/foo/bar/cmd/baz', 'github.com/foo/bar/cmd/baz')`, or more likely using some Skylark list comprehension so I can just maintain a list of known import paths and their `deps` for autobuilding goodness. Does it seem like that's something Tilt can support?

It might even be cool to have the signature of `ko_build` take a map of importpath -> deps, and generate the rest.

Another thing we might be able to optimize a bit more is the repeated calls to `ko publish`. If there are multiple images to publish, we can probably make them quicker if we batch them up. I recently added [`remote.MultiWrite`](https://godoc.org/github.com/google/go-containerregistry/pkg/v1/remote#MultiWrite) to go-containerregistry, which deduplicates shared layer uploads, reuses auth handshakes, and uploads blobs and manifests in parallel, and I'd like to use it to speed up `ko resolve` (https://github.com/google/ko/issues/236) -- it'd be great to be able to leverage that optimization in this extension too, if that's possible. I'd be willing to support a  `ko publish [importpath1] [importpath2] ...` mode if that would help here.

nicks:
@ImJasonH thanks! ya, let me try to break out a few nice ideas from your comment:

- re: `ko://` - Unforunately, if you do that right now, you get:

```
Error: Argument 1 (ref): can't parse "ko://github.com/mattmoor/warm-image/cmd/sleeper": invalid reference format
```

Tilt requires all image references to validate against this reference grammar: https://github.com/docker/distribution/blob/master/reference/reference.go#L4

I'd be open to changing Tilt to use a different reference parser here. is there one that supports the ko format?

- re: ko publish multiplexing - this is hard for us to do right now : (. custom_build isn't synchronous - it adds the shell script you give it to a planner, and the planner tries to parallelize things when possible. Let me think on this a bit...

- re: ko publish vs ko publish --local - right now this extension uses ko publish --local, and then has Tilt push the image. Do you think it would be significantly faster to have Ko publish the image to the registry (because we'd get to skip the local docker image store)? I might try taking a crack at that.

as an aside, I lol'd at some of the docs around getting Ko to work with Kind/Minikube and multi-level repository names...we ran into the same issues...it might be helpful to join forces on standard decision trees for this; I've been trying to push https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/generic/1755-communicating-a-local-registry but it's been slow going...

ko/README.md
```diff
@@ -0,0 +1,23 @@
+# Ko
+
+Author: [Nick Santos](https://github.com/nicks)
+
+Use Ko (https://github.com/google/ko) to build images for Tilt.
+
+## Functions
+
+### `ko_build(ref: str, import_path: str, deps: List[str])`
```
landism, line 9:
```suggestion
### `ko_build(ref: str, import_path: str, deps: List[str] = [])`
```

ko/Tiltfile
```diff
@@ -0,0 +1,30 @@
...
+  ]
+
+  if not import_path.startswith("."):
+    # Ko needs the source code to be checked out.
+    # https://github.com/google/ko/issues/237
+    commands += ["go get {import_path}".format(import_path=import_path)]
+
+  commands += [
+    "IMAGE=$(ko publish --local {import_path})".format(import_path=import_path),
+    'if [[ "$?" != "0" ]]; then exit 1; fi',
```
landism, line 23: Doesn't hurt, but I think the `set -e` renders this redundant

ko/Tiltfile
```diff
@@ -0,0 +1,30 @@
...
+    "docker tag $IMAGE $EXPECTED_REF",
+  ]
+  custom_build(
+    ref=ref,
+    command=["bash", "-c", ";\n".join(commands)],
+    deps=deps + [__file__],
```
landism, line 29: Just to confirm my understanding, we only need the `__file__` here to catch iterations on the extension itself that don't affect any of the custom_build args, right?

landism (approved):
I hadn't looked at `ko` before today. After reading the docs, it seems like the idea is that you provide yaml that specifies images as ko://gopath and it takes care of building and publishing and substituting images.

Instead of exposing `ko_build`, would it be more in line with the main `ko` use case to make a `ko_yaml` that digs through the yaml looking for `ko://` references, replaces them with non-`ko://` references, and declares `ko publish` `custom_build`s for those refs? (which would also get around the ref parsing issue)

ko/Tiltfile
nicks: no, this is a workaround for https://github.com/tilt-dev/tilt/issues/3891, left a comment about this

ko/Tiltfile
nicks: oops, good catch, ya, this is vestigial from an earlier version

ko/README.md
nicks: done

nicks:
re: a ko_yaml function - ya, i like that, let me try in a subsequent PR

PR #102, opened 2020-11-10
Add extension for current_namespace

ajkamel:
This can be used to automate your Tiltfile dynamically by fetching the
current namespace and using it within your Tiltfile using string
templates.

All you would have to do then would be to change your namespace using
something like `kubens "your-namespace"` before running `tilt up`.

@victorwuky 
@shaimendel


victorwuky:
@nicks would you please be able to review this one? Thanks.

nicks (approved):
LGTM!

PR #104, opened 2020-11-19
configmap: add an extension for auto-deploying configmaps from files

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/configmap:

22504eb3ff76c652a45a8b0cc6d8d153fde110ed (2020-11-19 17:34:20 -0500)
configmap: add an extension for auto-deploying configmaps from files


nicks:
i found this helpful when iterating on some grafana stuff

configmap/README.md
```diff
@@ -0,0 +1,44 @@
...
+### configmap_create
+
+```
+configmap_create(name: str, namespace: str = "", from_file: Union[str, List] = None, watch: bool = True)
+```
+
+Deploys a config map. Equivalent to
+
+```
+load('ext://configmap', 'configmap_yaml')
```
landism, line 29: maybe I'm being nitty - feels slightly weird to have this in the block and say "equivalent to" when these functions don't have the side effects of `load`, and seems like it'd be just as clear without the `load`?

configmap/README.md
```diff
@@ -0,0 +1,44 @@
...
+Author: [Nick Santos](https://github.com/nicks)
+
+Helper functions for creating Kubernetes configmaps.
+
+## Functions
+
+### configmap_yaml
+
+```
+configmap_yaml(name: str, namespace: str = "", from_file: Union[str, List] = None, watch: bool = True): Blob
```
landism, line 12:
```suggestion
configmap_yaml(name: str, namespace: str = "", from_file: Union[str, List[str]] = None, watch: bool = True): Blob
```

configmap/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+Returns YAML for a config map generated from a file.
+
+* `from_file` ( str ) – equivalent to `kubectl create configmap --from-file`
+* `watch` ( bool ) - auto-reload if the files change
+
+### configmap_create
+
+```
+configmap_create(name: str, namespace: str = "", from_file: Union[str, List] = None, watch: bool = True)
```
landism, line 23:
```suggestion
configmap_create(name: str, namespace: str = "", from_file: Union[str, List[str]] = None, watch: bool = True)
```

configmap/Tiltfile
```diff
@@ -0,0 +1,63 @@
...
+    "configmap",
+    name,
+  ]
+
+  if namespace:
+    args.extend(["-n", namespace])
+
+  generator = False
+  if from_file:
+    if type(from_file) == "string":
```
landism, line 30: ffti, but a pattern I've liked for this:

```
if type(from_file) == "string":
  from_file = [from_file]
if type(from_file) == "list":
  # this block now handles both cases
else:
  # error
```

landism (approved):

configmap/README.md
nicks: done

configmap/README.md
nicks: done

configmap/README.md
nicks: done

configmap/Tiltfile
nicks: i like that! done

PR #106, opened 2020-11-30
add kubebuilder extension

ellenkorbes:
@nicks Can you double check the Tiltfile? Line 26 especially feels a bit embarrassing, but I couldn't find the proper way.

Video will follow after this is approved.

### DIFF
--- a/kubebuilder/Tiltfile
+++ b/kubebuilder/Tiltfile
@@ -1,0 +1,60 @@
+load('ext://restart_process', 'docker_build_with_restart')
+
+def kubebuilder(DOMAIN, GROUP, VERSION, KIND, IMG='controller:latest'):
+
+    DOCKERFILE = '''FROM golang:alpine
+    WORKDIR /
+    COPY ./bin/manager /
+    CMD ["/manager"]
+    '''
+
+    def yaml():
+        return local('cd config/manager; kustomize edit set image controller=' + IMG + '; cd ../..; kustomize build config/default')
+    
+    def manifests():
+        return 'controller-gen crd:trivialVersions=true rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases;'
+    
+    def generate():
+        return 'controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./...";'
+    
+    def vetfmt():
+        return 'go vet ./...; go fmt ./...'
+    
+    def binary():
+        return 'CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o bin/manager main.go'
+
+    installed = local("which kubebuilder")
+    print("kubebuilder is present:", installed)
+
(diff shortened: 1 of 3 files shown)


kubebuilder/Tiltfile
```diff
@@ -0,0 +1,64 @@
...
+        local("kubebuilder create api --resource --controller --group %s --version %s --kind %s" % (GROUP, VERSION, KIND))
+    
+    local(manifests() + generate())
+    
+    local_resource('CRD', manifests() + 'kustomize build config/crd | kubectl apply -f -', deps=["api"])
+    
+    k8s_yaml(yaml())
+    
+    deps = ['controllers', 'main.go']
+    if initting == False:
```
nicks, line 50: i'm not sure I understand this if branch? can you leave a comment on what it's for?

kubebuilder/Tiltfile
```diff
@@ -0,0 +1,64 @@
...
+        deps.append('api')
+    
+    local_resource('Watch/Compile', generate() + binary(), deps=deps, ignore=['bin', '*/*/zz_generated.deepcopy.go'])
```
nicks, line 53: not sure i understand why you need the ignore=['bin'] if you're not watching this directory?

kubebuilder/Tiltfile
```diff
@@ -0,0 +1,64 @@
...
+        initting = True
+        local("kubebuilder init --domain %s" % DOMAIN)
+    
+    if os.path.exists('api') == False:
+        initting = True
+        local("kubebuilder create api --resource --controller --group %s --version %s --kind %s" % (GROUP, VERSION, KIND))
+    
+    local(manifests() + generate())
+    
+    local_resource('CRD', manifests() + 'kustomize build config/crd | kubectl apply -f -', deps=["api"])
```
nicks, line 45: *sigh* i wish we had a better answer for this

nicks (approved):

kubebuilder/Tiltfile
ellenkorbes: Same as above; removed.

kubebuilder/Tiltfile
ellenkorbes: This was old stuff from when me and Dan Miller worked on it; I don't know either. Removed now and everything seems to work fine still. 🤷

kubebuilder/Tiltfile
ellenkorbes: WDYM?

nicks: tilt can't track resources that are applied with `kubectl apply`. But the primitives for dynamically generating YAML also aren't great. There's some discussion on it here:
https://github.com/tilt-dev/tilt/issues/2664
https://github.com/tilt-dev/tilt/issues/3046

DM and I had spent a bunch of time thinking about what a better solution to this might be...it's a problem that a lot of build systems wrestle with...

PR #108, opened 2020-12-02
add support for tag references

KOGI:
This allows specifying a tag name as the resource branch. Tags are a special case, and things break if you try to do `git checkout origin/tagName` or `git checkout origin/tags/tagName`, so we must rewrite to `git checkout tags/tagName`

@nicks 
@victorwuky

### DIFF
--- a/git_resource/Tiltfile
+++ b/git_resource/Tiltfile
@@ -48,7 +48,10 @@ def git_checkout(repository_url, checkout_dir=None, unsafe_mode=False):
                     has_local_changes = int(str(local('cd %s && git status -sbuno | wc -l' % checkout_dir, quiet=True)).strip()) > 1  # result greater than 1 indicates local modifications are present
 
                 if unsafe_mode or not has_local_changes:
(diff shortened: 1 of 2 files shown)


git_resource/Tiltfile
```diff
@@ -48,7 +48,10 @@ def git_checkout(repository_url, checkout_dir=None, unsafe_mode=False):
                     has_local_changes = int(str(local('cd %s && git status -sbuno | wc -l' % checkout_dir, quiet=True)).strip()) > 1  # result greater than 1 indicates local modifications are present
 
                 if unsafe_mode or not has_local_changes:
-                    local('cd %s && git fetch -a origin && git checkout -f origin/%s && git submodule update --init' % (checkout_dir, branch), quiet=True)
+                    if branch[:5] != 'tags/' and branch[:7] != 'origin/':
```
nicks, line 51: fwiw, branch.startswith would also work!
https://docs.bazel.build/versions/master/skylark/lib/string.html#startswith

nicks (approved):
lgtm!

git_resource/Tiltfile
KOGI: Ah, I forgot that was available! Thank you!

PR #110, opened 2020-12-04
syncback: first pass at syncback extension

maiamcc:
Before I do too much polish/documentation/the extra configuration stuff that will make this ext actually usable, wanted a gut check that this is plausibly the right way to go. WDYT @nicks @landism ?

[Here's this extension in action on the sample rails app](https://github.com/tilt-dev/tilt-example-rails/tree/syncback-extension/syncback). Experiment with adding new gems/node modules/controllers and then running `syncback`.


syncback/Tiltfile
```diff
@@ -0,0 +1,26 @@
+# -*- mode: Python -*-
+
+krsync_path = os.path.join(os.getcwd(), 'krsync.sh')
+
+
+def syncback(name, deploy, paths, target_dir='.'):
+    # name (str) = resource name
+    # k8s_identifier (str) = thing to exec on (pod, deployment/blah, etc)
```
nicks, line 8: hmmmm....i'm confused...certain parts of this extension say that k8s_identifier can be a pod, but in krsync.sh it says it has to be a deployment?

syncback/Tiltfile
```diff
@@ -0,0 +1,26 @@
...
+    # paths (List[str]) = paths ON CONTAINER to sync. May be files or dirs. May be absolute, or releative to working dir of container
+    # target_dir (str) = dir on local FS to copy files back to (maybe abs or relative to Tiltfile)
+
+    # TODO: if you're rsync-savvy you might want to do the wildcarding manually--
+    #   give an option to turn off automatic +'***'
+    includes = ' '.join(['--include="{}***"'.format(p) for p in paths])
```
nicks, line 14: what does it mean to syncback multiple paths to a single directory? can you give an example?

syncback/krsync.sh
```diff
@@ -0,0 +1,30 @@
...
+    shift
+fi
+
+# ensure our rsync bin is present on container in the expected location
+## TODO: option for windows bin, depending on container OS
+tar_path=${0%/*}/rsync.tilt.tar
+cat $tar_path | kubectl $namespace exec -i deployment/$depl -- /bin/sh -c 'ls rsync.tilt 2>/dev/null && cat>/dev/null || tar -xf -' > /dev/null
+
+exec kubectl $namespace exec -i deployment/$depl -- "$@"
+
```
nicks, line 30: did you intend to commit syncback/rsync.tilt?

syncback/krsync.sh
```diff
@@ -0,0 +1,30 @@
...
+    depl=$1
+    shift
+    namespace="-n $1"
+    shift
+fi
+
+# ensure our rsync bin is present on container in the expected location
+## TODO: option for windows bin, depending on container OS
+tar_path=${0%/*}/rsync.tilt.tar
+cat $tar_path | kubectl $namespace exec -i deployment/$depl -- /bin/sh -c 'ls rsync.tilt 2>/dev/null && cat>/dev/null || tar -xf -' > /dev/null
```
nicks, line 27: have you considered doing separate kubectl calls - one to check if the file exists, and then use `kubectl cp`?

the way this is written, you need to copy 2MB across the network, even if the file is already there

nicks, line 27: also, there's a slight bug where this only works if the working directory of the container is /app

nicks (commented):
cool! all of my comments are easily fixable things

syncback/Tiltfile
maiamcc: if i have `paths=['package.json', 'yarn.lock', 'migrations/']` and directory `'./letters'` then the resulting files would be `letters/package.json`, `letters/yarn.lock`, `letters/migrations/`. Does that answer your q?

syncback/krsync.sh
maiamcc: ah whoops i guess that's not necessary is it?

syncback/krsync.sh
maiamcc: good call, will do it that way

maiamcc: wait how? i don't follow.

maiamcc: (though this may be moot because one of the things i'm going to do it put `rsync.tilt` somewhere outside of the working dir)

nicks: the --rsync-path flag above?

syncback/Tiltfile
maiamcc: whoops, this comment is out of date. yes it has to be a deployment. (how important do you think it is that people be able to pass a different identifier--e.g. a pod, job name, etc.?)

syncback/krsync.sh
maiamcc: aha whoops yes!

syncback/Tiltfile
nicks: ya, jobs and statefulsets are pretty popular. is it hard to require the resource name? like "deployment/my-deployment"

fwiw, i found the code that kubectl uses to find the pod for an exec recently! https://github.com/kubernetes/kubectl/blob/master/pkg/polymorphichelpers/helpers.go#L84

maiamcc:
@nicks fixed up all the stuff you commented on. lmk if anything looks glaringly weird, otherwise i'll merge this and do documentation and other TODO's in separate PRs

syncback/krsync.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+# Adapted from Karl Bunch's post on Server Fault:
+# https://serverfault.com/questions/741670/rsync-files-to-a-kubernetes-pod/887402#887402
+
+set -e
+
+# make sure local tmp dir exists
+[[ -d /tmp/rsync.tilt ]] || mkdir /tmp/rsync.tilt
```
nicks, line 9: the more idiomatic way to do this is

```
mkdir -p /tmp/rsync.tilt
```

and not bother checking

nicks (approved):

syncback/Tiltfile
```diff
@@ -0,0 +1,30 @@
...
+
+def syncback(name, k8s_object, paths, src_dir, target_dir='.', namespace=''):
+    # name (str) = resource name
+    # k8s_object (str) = thing to exec on (pod/foo, deployment/bar, etc)
+    # paths (List[str]) = paths ON CONTAINER to sync. May be files or dirs. May be absolute, or relative to src_dir
+    # src_dir = dir ON CONTAINER that paths are relative to
+    # target_dir (str) = dir on local FS to copy files back to (maybe abs or relative to Tiltfile)
+
+    if not src_dir.endswith('/'):
+        fail('src_dir must be a directory and have a trailing slash')
```
maiamcc, line 14: @nicks does this seem like an ok approach (as opposed to silently tacking on a trailing slash, as opposed to just passing in the value the user gave us)? doing this b/c trailing slash is semantic in rsync, and i don't expect the average tilt user to get that nuance

djcp: having done a bunch of work with rsync in the past, this makes total sense to me.

syncback/Tiltfile
```diff
@@ -0,0 +1,30 @@
...
+
+    # TODO: if you're rsync-savvy you might want to do the wildcarding manually--
+    #   give an option to turn off automatic +'***'
+    includes = ' '.join(['--include="{}***"'.format(p) for p in paths])
+
+    remote_name = 'dummy'
+    if namespace:
+        remote_name = 'dummy@{}'.format(namespace)
+    local_resource(name, '{krsync} {obj} -avO --progress --stats --delete -T=/tmp/rsync.tilt {includes} --exclude="*" {remote}:{src} {target}'.
```
djcp, line 23: To avoid possible collisions, could you have the tempdir use `mktemp` and pass that value to krsync?

syncback/Tiltfile
```diff
@@ -0,0 +1,30 @@
+# -*- mode: Python -*-
+
+krsync_path = os.path.join(os.getcwd(), 'krsync.sh')
+
+
+def syncback(name, k8s_object, paths, src_dir, target_dir='.', namespace=''):
+    # name (str) = resource name
+    # k8s_object (str) = thing to exec on (pod/foo, deployment/bar, etc)
+    # paths (List[str]) = paths ON CONTAINER to sync. May be files or dirs. May be absolute, or relative to src_dir
```
djcp, line 9: What path would you use if you wanted everything under the root to be synced back if it's changed in-container, and does that work OK?

maiamcc: good call, this is a use case i hadn't considered. EXTREMELY counterintuitively, the path `''` will get everything in `src_dir`. Let me change it to make `paths` optional, and passing NO paths defaults to pulling everything from `src_dir`.

maiamcc: ☝️ does that sound like reasonable behavior?

djcp: Yeah, that sounds reasonable.

PR #114, opened 2020-12-10
syncback: default excludes, can pass excludes

maiamcc:
Flagged by @djcp (🙏🏻):
> One issue with this: if you set paths to include the entire repo (so empty), this'll remove your `.git/` directory in the host because of the `--delete` directive and how it interacts with default ignores.  Probably an explicit additional `excludes` option could fix this.

Not convinced that this API (`paths` is NON-optional, pass `[]` to watch entire dir) is the right API, but I wanted to make it a little harder to watch the entire dir/make sure it's a deliberate choice when it happens, b/c it's a dangerous option and you need to set your `ignore` values correctly or you risk blowing away data. Not quite sure how to do this tho, thoughts?

(Also, called the parameter `ignore` to match `docker_build` params, but happy for it to be whatever--`ignores`? `excludes`?)

### DIFF
--- a/syncback/README.md
+++ b/syncback/README.md
@@ -26,37 +26,42 @@ For every syncback resource you want to create, invoke `syncback` with the follo
 
 * **name (str)**: name of the created local resource
 * **k8s_object (str)**: a Kubernetes object identifier (e.g. `deploy/my-deploy`, `job/my-job`, or a pod ID) that Tilt can use to select a pod. As per the behavior of `kubectl exec`, we will act on the first pod of the specified object, using the first container by default
+* **paths (List[str])**: paths *in the remote container* to sync, relative to `src_dir`. May be files or directories. Note that these must not begin with `./`. Pass an empty list to sync all of src_dir. THIS OPTION RISKS WIPING OUT FILES LOCALLY if the files exist at `target_dir` but not in the container. Tilt will protect some files automatically, but you will probably want to use the `ignores` parameter to explicitly protect other files that exist locally but not on the container.
 * **src_dir (str)**: directory *in the remote container* to sync from. Any `paths`, if specified, should be relative to this dir. This path *must* be a directory and must contain a trailing slash (e.g. `/app/` is acceptable; `/app` is not)
 
 You may also pass the following optional parameters:
-* **paths (List[str], optional)**: paths *in the remote container* to sync, relative to `src_dir`. May be files or directories. Note that these must not begin with `./`. If no paths are provided, we sync the entire `src_dir`
+* **ignore (List[str], optional)**: files to ignore when syncing (relative to src_dir).
 * **target_dir (str, optional)**: directory *on the local filesystem* to sync to. Defaults to `'.'`
-* **container (str, optiona)**: name of the container to sync from (by default, the first container)
+* **container (str, optional)**: name of the container to sync from (by default, the first container)
 * **namespace (str, optiona)**: namespace of the desired `k8s_object`, if not `default`.
 * **verbose (bool, optional)**: if true, print additional rsync information.
 
-### Example invocations:
-1. Create a local resource called "syncback-js" which connects to the first pod of "deploy/frontend" (and the default container) and syncs "/app/package.json" and "/app/yarn.lock" to local directory "./frontend".
+### Example invocations
+1. Create a local resource called "syncback-js" which connects to the first pod of "deploy/frontend" (and the default container) and syncs "/app/package.json" and "/app/yarn.lock" to local directory "./frontend":
     ```python
-    syncback('syncback-js', 'deploy/frontend', '/app/',
+    syncback('syncback-js', 'deploy/frontend',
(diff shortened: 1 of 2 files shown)


syncback/README.md
```diff
@@ -26,37 +26,42 @@ For every syncback resource you want to create, invoke `syncback` with the follo
 
 * **name (str)**: name of the created local resource
 * **k8s_object (str)**: a Kubernetes object identifier (e.g. `deploy/my-deploy`, `job/my-job`, or a pod ID) that Tilt can use to select a pod. As per the behavior of `kubectl exec`, we will act on the first pod of the specified object, using the first container by default
+* **paths (List[str])**: paths *in the remote container* to sync, relative to `src_dir`. May be files or directories. Note that these must not begin with `./`. Pass an empty list to sync all of src_dir. THIS OPTION RISKS WIPING OUT FILES LOCALLY if the files exist at `target_dir` but not in the container. Tilt will protect some files automatically, but you will probably want to use the `ignores` parameter to explicitly protect other files that exist locally but not on the container.
```
nicks, line 29: > Note that these must not begin with `./`

what's the reasoning behind this restriction? would it make sense to add code to check for this?

syncback/Tiltfile
```diff
@@ -30,11 +36,22 @@ def syncback(name, k8s_object, src_dir, paths=None, target_dir='.', container=''
...
     if paths:
         # TODO: if you're rsync-savvy you might want to do the wildcarding manually--
         #   give an option to turn off automatic +'***'
         includes = ' '.join(['--include="{}***"'.format(p) for p in paths])
         incl_excl = '{} --exclude="*"'.format(includes)
+    else:
+        # Sync the entire src_dir. Danger, Will Robinson! Ignore some stuff
+        # that probably exists locally but not in your container so it
+        # doesn't get wiped out locally on your first sync
+        ignore.extend(DEFAULT_IGNORES)
```
nicks, line 51: it's generally considered bad practice to modify your inputs. for example, if you have

```
shared_ignores = []
syncback(..., ignore = shared_ignores)
syncback(..., ignore = shared_ignores)
```

you will end up modifying it twice.

maiamcc: whooops I guess I assumed that starlark would guard against that. Will change tomorrow.

syncback/README.md
maiamcc: rsync doesn't like that syntax: it won't end up being evaluated relative to src_dir (it doesn't get rejected outright, idk what it does instead... maybe literally looks for a directory called `.`? but it doesn't translate to `$pwd` in rsync).

Yep, good call, i'll put that check in the code tomoro

maiamcc:
@nicks fixed both of your callouts, and tweaked the API again--on a suggestion from @DCP added a way to configure whether rsync is run with `--delete` (default: False). Now that there's some more control over deletion behavior and less of a risk that the user will accidentally blow away files, I made paths back into an optional arg. (Instead of passing `paths=[]` to sync the whole container, just omit the paths arg.) wdyt/

syncback/Tiltfile
```diff
@@ -20,25 +22,56 @@ def syncback(name, k8s_object, src_dir, paths=None, target_dir='.', container=''
...
     if paths:
         # TODO: if you're rsync-savvy you might want to do the wildcarding manually--
         #   give an option to turn off automatic +'***'
         includes = ' '.join(['--include="{}***"'.format(p) for p in paths])
         incl_excl = '{} --exclude="*"'.format(includes)
+    else:
+        # Sync the entire src_dir. Danger, Will Robinson! Exclude some stuff
+        # that may exist locally but not in your container so it
+        # doesn't get wiped out locally on your first sync
+        to_exclude.extend(DEFAULT_EXCLUDES)
```
nicks, line 66: i think this still has the same problem - it mutates an argument. probably simpler to reason about if you simply don't use mutation (e.g., do to_exclude = DEFAULT_EXCLUDES + to_exclude)

syncback/Tiltfile
```diff
@@ -9,7 +9,9 @@ verify_rsync_path = os.path.join(os.getcwd(), 'verify_rsync.sh')
 print('-- syncback extension checking for local rsync --')
 local('{} {}'.format(verify_rsync_path, MIN_LOCAL_RSYNC_VERSION))
 
-def syncback(name, k8s_object, src_dir, paths=None, target_dir='.', container='', namespace='', verbose=False):
+DEFAULT_EXCLUDES = ['.git', '.gitignore', '.dockerignore', 'Dockerfile', '.tiltignore', 'Tiltfile', 'tilt_modules']
```
nicks, line 12: this is probably OK for now, but it probably won't work if you have sub-Tiltfiles. You probably need to make them relative to `config.main_dir`

nicks (approved):

syncback/Tiltfile
maiamcc: These are paths on the container though / relative to src_dir, so I think it'll be fine!

Issue #116, opened 2020-12-19
Infinite loop with helm_remote


VeryStrongFingers:
I've been running into an issue where Tilt is rebuilding on a loop, invoked from helm chart content changes

It is reproducible with the Tiltfile contents being:
```
load('ext://helm_remote', 'helm_remote')
helm_remote('traefik', repo_url='https://helm.traefik.io/traefik', values=['traefik-helm-values.yaml'])
```
now create a file called `traefik-helm-values.yaml` in the same directory (empty file contents, specific values don't matter)

Run `tilt up`, and it handles it all as expected, now make a change to the `traefik-helm-values.yaml` (any change, eg. add an `#`)
Tilt will then be on an indefinite loop with the following output

```
1 File Changed: [traefik-helm-values.yaml] • (Tiltfile)
... (usual output)

32 Files Changed: [.helm/traefik/latest/traefik .helm/traefik/latest/traefik/.helmignore .helm/traefik/latest/traefik/Chart.yaml .helm/traefik/latest/traefik/Guidelines.md .helm/traefik/latest/traefik/LICENSE ...] • (Tiltfile)
Beginning Tiltfile execution
local: helm repo add traefik https://helm.traefik.io/traefik || helm repo add traefik https://helm.traefik.io/traefik --force-update
 → "traefik" has been added to your repositories
local: rm -rf /home/user/helm-tilt-issue/.helm/traefik/latest/traefik
 → [no output]
local: helm pull traefik/traefik --untar --destination /home/user/helm-tilt-issue/.helm/traefik/latest
 → [no output]
local: grep --include='*.yaml' --include='*.yml' -rEil '\bkind[^\w]+CustomResourceDefinition\s*$' /home/user/helm-tilt-issue/.helm/traefik/latest/traefik || exit 0
 → /home/user/helm-tilt-issue/.helm/traefik/latest/traefik/crds/tlsstores.yaml
 → /home/user/helm-tilt-issue/.helm/traefik/latest/traefik/crds/traefikservices.yaml
 → /home/user/helm-tilt-issue/.helm/traefik/latest/traefik/crds/ingressrouteudp.yaml
 → /home/user/helm-tilt-issue/.helm/traefik/latest/traefik/crds/middlewares.yaml
 → /home/user/helm-tilt-issue/.helm/traefik/latest/traefik/crds/ingressroute.yaml
[...]

PR #117, opened 2020-12-19
helm_remote fix endless chart update loop + security improvements

VeryStrongFingers:
Refactored helm_remote Tiltfile to remove the pessimism, and only perform changes when required
This stops the endless looping issue without having to change your `.tiltiginore`
resolves (my reported) issue https://github.com/tilt-dev/tilt-extensions/issues/116 and contains other improvements

also noticed a few spots for improving security, namely avoiding directory traversal from arbitrary inputs, especially because `rm` is involved with this.
This prevents chart/repo names like `something/../../` or `something && cat /etc/passwd` from being entertained, and shell 

**Changes** (there's no breaking changes)
- Fixed the indefinite loop that was possible due to how remote helm charts were being handled 
- Chart will only be re-downloaded to filesystem (`helm pull`) when required
  - this is based on a version comparison between local cached (if existent) and remote chart version
- Repository update is done independently
- Improved security for many input arguments that derive from user input
   - Added validation and failure messages for unsuitable input values
- Removed the need for `--force-update` which divides users depending on their helm version
  - (helm repo update resolves that need)
- Reduced repeated code (ie. command argument parsing for auth/user)

### DIFF
--- a/helm_remote/Tiltfile
+++ b/helm_remote/Tiltfile
@@ -40,7 +40,8 @@ metadata:
 def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], set=[], namespace='', version='', username='', password='', allow_duplicates=False, create_namespace=False):
     # ======== Helper methods
     def get_local_repo(repo_name, repo_url):
-        added_helm_repos = decode_yaml(local('helm repo list --output yaml'))
+        # if no repos are present, helm exit code is >0 and stderr output buffered
+        added_helm_repos = decode_yaml(local('helm repo list --output yaml 2>/dev/null || true', command_bat='helm repo list --output yaml 2>&1 || ver>nul'))
         repo = [item for item in added_helm_repos if item['name'] == chart and item['url'] == repo_url]
 
         return repo[0] if len(repo) > 0 else None
--- a/helm_remote/Tiltfile
+++ b/helm_remote/Tiltfile
@@ -78,6 +78,24 @@ def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], se
     if version == '':
         version = 'latest'
 
+    # ======== Validate before we start trusting chart/repo names
+
+    # Based on helm chart conventions, and the fact we don't want anyone traversing directories
+    # validate is to essentially ensure there's no special characters aside from '-' being used
+    # https://helm.sh/docs/chart_best_practices/conventions/#chart-names
+
+    if chart.replace('-', '').isalnum() == False or chart == chart.replace('.', ''):
+        # https://helm.sh/docs/chart_best_practices/conventions/#chart-names
+        fail('Chart name is not validate')
+
+    if repo_name != chart and (repo_name.replace('-', '').isalnum() == False or repo_name == repo_name.replace('.', '')):
+        # https://helm.sh/docs/chart_best_practices/conventions/#chart-names
+        fail('Repo name is not validate')
+
+    if version != 'latest' and version != version.replace('/', '').replace('\\', ''):
+        fail('Version cannot contain a forward slash')
+
+
     # ======== Determine state of existing helm repo
     if repo_url != '':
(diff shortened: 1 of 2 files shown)


helm_remote/Tiltfile
```diff
@@ -71,29 +119,36 @@ def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], se
...
+        pull_command = 'pull %s/%s --untar --destination %s' % (repo_name, chart, pull_target)
+
+        # ======== Perform Installation
+        if cached_chart_exists == True:
+            local('rm -rf %s' % chart_target,
+                  command_bat='if exist %s ( rd /s /q %s )' % (chart_target, chart_target))
+
+        local(build_helm_command(pull_command, (username, password), version))
+
+        install_crds(chart, chart_target)
```
nicks, line 151: the install_crds command needs to be outside the if block, right?

nicks (commented):
thanks for contributing!

This looks pretty good except for the one bug i noticed below

Also cc @KOGI (the original author of this extension) if he has any other validation that's important to check here

helm_remote/Tiltfile
```diff
@@ -38,31 +38,79 @@ metadata:
...
+    # ======== Validate before we start trusting chart/repo names
+
+    # Based on helm chart conventions, and the fact we don't want anyone traversing directories
+    # validate is to essentially ensure there's no special characters aside from '-' being used
+    # str.isalnum accepts dots, which is only dangerous when slashes are allowed 
+    # https://helm.sh/docs/chart_best_practices/conventions/#chart-names
+
+    if chart.replace('-', '').isalnum() == False or chart != chart.replace('.', ''):
+        # https://helm.sh/docs/chart_best_practices/conventions/#chart-names
+        fail('Chart name is not validate')
```
KOGI, line 91: Should be `Chart name is not valid` or `Chart name is invalid`

helm_remote/Tiltfile
```diff
@@ -38,31 +38,79 @@ metadata:
...
+        local_helm_repo = get_local_repo(repo_name, repo_url)
+
+        if (local_helm_repo == None):
+            # Unaware of repo, add it 
+            repo_command = 'repo add %s %s' % (shlex.quote(repo_name), shlex.quote(repo_url))
+        else:
+            # Helm is already aware of the chart, update repo (unfortunately you cannot specify a single repo)
+            repo_command = 'repo update'
+
+        # Add authentification for adding the repository if credentials are provided
```
KOGI, line 112: `s/authentification/authentication`

KOGI (changes requested):

helm_remote/Tiltfile
VeryStrongFingers: @nicks I left it purposefully in the `pull` scope because the result of `install_crds()` would be idempotent, when the helm version version has not changed.

I'm happy to move it outside this scope though if it's preferred

nicks: Hmmm...I think you're mistaken. The Helm chart is stored locally. The CRDs are stored in the cluster. If you reset the cluster, the Helm version will be the same, but the CRDs will need to be installed still.

helm_remote/Tiltfile
VeryStrongFingers: can't believe I made a new word 😭

helm_remote/Tiltfile
VeryStrongFingers: That's a really valid point, it's not checking the helm chart metadata, and it's possible for two different Tiltfiles to use the same chart, so it's a possible case.
I've moved it outside this scope now

VeryStrongFingers:
@KOGI @nicks I've done all requested changes, let me know if there's anything further

nicks (approved):
"authentification" is a great word, TBH

LGTM!

PR #118, opened 2020-12-23
Tidy up extension, add support for trigger modes and info only

mattj-io:
Adds support for different trigger modes, and a switch to control the exit code for gating or information only. Tidy up arg names, update README

### DIFF
--- a/snyk/README.md
+++ b/snyk/README.md
@@ -1,43 +1,59 @@
 # Use Snyk to test your configuration files
 
 Author: [Shai Mendel](https://github.com/shaimendel)
-* Additional contributor: [Jim Armstrong](https://github.com/jimcodified)
+* Additional contributors: 
+	[Jim Armstrong](https://github.com/jimcodified)
+	[Matt Jarvis](https://github.com/mattj-io)
 
 [Snyk](https://https://snyk.io/) is a product to help developers to continuously find and fix vulnerabilities in open source libraries, containers and Infrastructure as Code configurations.
 
+## Requirements
+
+* The `snyk` binary must be on your path. See https://support.snyk.io/hc/en-us/articles/360003812538-Install-the-Snyk-CLI for details on installing, and register for a free Snyk account at https://app.snyk.io/login
+
 ## Usage
 
-The `snyk` extension can be run like so:
+The extension takes a number of arguments:
+
+name: a name for the resource. Useful for labeling when running multiple tests. Defaults to 'snyk'
+target: path to the file/dir or container image name:tag to test
+test_type: one of 'oss', 'container', or 'iac' corresponding to the various Snyk CLI tests
+test_deps: automatically set for test types other than container, allows for setting an external file/dir dependency
+extra_opts: additional CLI options; appended to the snyk command. see `snyk --help` for usage.
+          - when using the container switch this is where to insert the '--file=path/to/Dockerfile' option
+trigger: trigger mode - auto or manual, defaults to manual
+mode: control Snyk exit code, info will always exit 0 - gate or info, defaults to gate
+
+The `snyk` extension can be run in a variety of different ways:
 
 ```python
+# Load the extension
 load('ext://snyk', 'snyk')
 
-docker_build('myimage:v1','.')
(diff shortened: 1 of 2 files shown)


snyk/README.md
```diff
@@ -1,43 +1,59 @@
...
+The extension takes a number of arguments:
+
+name: a name for the resource. Useful for labeling when running multiple tests. Defaults to 'snyk'
+target: path to the file/dir or container image name:tag to test
+test_type: one of 'oss', 'container', or 'iac' corresponding to the various Snyk CLI tests
+test_deps: automatically set for test types other than container, allows for setting an external file/dir dependency
+extra_opts: additional CLI options; appended to the snyk command. see `snyk --help` for usage.
+          - when using the container switch this is where to insert the '--file=path/to/Dockerfile' option
+trigger: trigger mode - auto or manual, defaults to manual
+mode: control Snyk exit code, info will always exit 0 - gate or info, defaults to gate
```
landism, line 25: "gate" feels a little weird here - it's not actually gating anything, right?
IIUC, snyk will run after the code has already potentially been built and deployed, and this just controls whether the snyk resource shows red or green in Tilt when snyk has found a problem.

And if it's not actually gating, I'm not sure under what situation I'd want to use `mode=='info'` - I could see wanting to semi-disable snyk while trying to get something to work. Is there a circumstance under which one would want to configure a Tiltfile so that a failing scan always turns up as green in Tilt and makes it easier to miss? I figure if the user is uninterested in scan results, they can just choose not to look at them.

None of this is blocking for this PR - I just wanted to check if this made sense.

snyk/README.md
```diff
@@ -1,43 +1,59 @@
...
-snyk('myimage:v1', 'container', 'snyk-cnr', 'example', '--file=Dockerfile')
-snyk('deployment.yaml', 'iac', 'snyk-iac')
+# Automatically scan Docker builds
+# Extract the immutable tag on build and write to a file in the filesystem
+custom_build('example-nodejs-image', 'docker build -t $EXPECTED_REF . && echo $EXPECTED_REF > /tmp/ref.txt', ['.'])
+# Extract the tag from the file and strip any newlines
+def get_tag():
+    return str(read_file('/tmp/ref.txt')).rstrip('\n')    
+# Run snyk automatically in informational mode when the file containing the tag changes
+snyk(get_tag(), 'container', 'snyk-cnr-auto', test_deps='/tmp/ref.txt', mode='info', extra_opts='--file=Dockerfile', trigger='auto')
```
landism, line 52: Feel free to ignore - a direction you might be interested in is a custom_build wrapper that deals with ref.txt calls both custom_build and snyk for the user, kind of like this:
https://github.com/tilt-dev/tilt-extensions/blob/4b695acaf25dfe906799e5473879d74f4a689fe6/restart_process/Tiltfile#L99

(that's an example of taking the custom_build args and some extra args, calling the underlying custom_build, and doing some other stuff)

landism (approved):
LGTM

Let me know if you want to get a review from Shai or Jim, or make any further changes, or if you'd like me to merge it as-is.

snyk/README.md
mattj-io: I had the same thoughts, but not sure the right language so open to suggestions. From a local workflow perspective, I want to know that something isn't right, but not sure I want a big red flag. If we exit non-zero the Tilt output is build failed, which feels a lot like psychologically gating right ?

mattj-io: @shaimendel is already aware of this PR, I think we're good for the minute

snyk/README.md
mattj-io: This looks interesting, will take a look :) From my perspective of wanting to run things locally before/after builds, the biggest things that would make life easier are access to primitives to do strong ordering in a Tiltfile, and access to internal variables like the build image tags. Other than that I'm loving Tilt !

mattj-io:
Hi folks, just checking if we can get this merged ? I'm writing a blog post about Tilt which is kind of blocked behind these changes ..

victorwuky:
@mattj-io : Going to merge as @landism is okay as is from previous comment.

PR #128, opened 2021-02-04
Configurable location of binary

pusewicz:
This change solves the issue of paths and other binaries conflicting with the name.

In my case, there's a RubyGem called tilt that also has the binary called tilt installed and was taking the precedence (because of the Ruby version manager).

Having the environment variable available, I can now have it configured and the check can be performed targeting the correct binary location.

### DIFF
--- a/min_tilt_version/Tiltfile
+++ b/min_tilt_version/Tiltfile
@@ -3,8 +3,8 @@ def min_tilt_version(min_version):
   # v0.13.0, built 2020-04-01
   # Also supports dev versions
   # v0.13.0-dev, built 2020-04-03
-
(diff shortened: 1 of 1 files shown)


nicks (approved):
thanks @pusewicz !

ya, we should probably have a way in tilt to get a reference for the tilt binary, but this seems like a good workaround for now

LGTM

nicks:
filed as https://github.com/tilt-dev/tilt/issues/4167

milas:
As of Tilt 0.19.6+, `sys.executable` can be used to reliably get the path to the Tilt binary that's running!

(See tilt-dev/tilt#4436)

PR #137, opened 2021-03-04
Initial creation of global_vars extension

KOGI:
@nicks


nicks (approved):
I like this solution! LGTM

PR #140, opened 2021-03-05
Prevent fetching updates multiple times if we're already up to date

KOGI:
@nicks 

As part of its process, helm_remote() calls `helm repo update` to be sure its using the latest version of the chart. However, if the project contains multiple helm_remote() calls for different charts in the same repository, it would call `helm repo update` EACH time. This was having significant performance impacts on the time required to process the initial Tiltfile.

### DIFF
--- a/helm_remote/Tiltfile
+++ b/helm_remote/Tiltfile
@@ -1,3 +1,21 @@
+load('ext://global_vars', 'get_global', 'set_global')
+
+
+def _get_skip():
+    return get_global(_get_skip_name())
+
+
+def _set_skip(value):
+    set_global(_get_skip_name(), value)
+
+
+def _get_skip_name():
+    return 'HELM_REMOTE_SKIP_UPDATES'
+
+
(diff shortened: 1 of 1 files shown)


helm_remote/Tiltfile
```diff
@@ -102,15 +120,19 @@ def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], se
     if repo_url != '':
         local_helm_repo = get_local_repo(repo_name, repo_url)
 
-        if (local_helm_repo == None):
-            # Unaware of repo, add it 
+        if local_helm_repo == None:
+            # Unaware of repo, add it
             repo_command = 'repo add %s %s' % (shlex.quote(repo_name), shlex.quote(repo_url))
             # Add authentication for adding the repository if credentials are provided
```
KOGI, line 126: lines 127-135 below are the real meat of this PR

nicks (approved):
nice! LGTM!

PR #141, opened 2021-03-08
Update the Cert Manager module version to 1.1.0 as default

davideimola:
Hi,
as requested on this following [issue](https://github.com/kubernetes-sigs/cluster-api/issues/4256#issuecomment-792630611), I have changed the default version for the cert-manager module.

I have tried to maintain the backward compatibility for version 0 too.

Furthermore, I have removed the `existCheck` because the [issue related](https://github.com/kubernetes-sigs/cluster-api/issues/4256#issuecomment-792630611) has been already fixed, so I think it is not required anymore!
Thanks

### DIFF
--- a/cert_manager/Tiltfile
+++ b/cert_manager/Tiltfile
@@ -4,15 +4,15 @@ kind: Namespace
 metadata:
   name: cert-manager-test
 ---
-apiVersion: cert-manager.io/v1alpha2
+apiVersion: cert-manager.io/{cert_manager_api_version}
 kind: Issuer
 metadata:
   name: test-selfsigned
   namespace: cert-manager-test
 spec:
-  selfSigned: {}
+  selfSigned: {{}}
 ---
-apiVersion: cert-manager.io/v1alpha2
+apiVersion: cert-manager.io/{cert_manager_api_version}
 kind: Certificate
 metadata:
   name: selfsigned-cert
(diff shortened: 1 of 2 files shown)


cert_manager/Tiltfile
```diff
@@ -26,27 +26,26 @@ spec:
 """
 
 # Deploys cert manager to your environment
-def deploy_cert_manager(registry="quay.io/jetstack", version="v0.16.1", load_to_kind=False, kind_cluster_name="kind"):
+def deploy_cert_manager(registry="quay.io/jetstack", version="v1.1.0", load_to_kind=False, kind_cluster_name="kind"):
     silent=True
+    if version.startswith('v0'):
```
fabriziopandini, line 31: This isn't super robust, might be folks from tilt can give some suggestion here...

cert_manager/Tiltfile
```diff
@@ -26,27 +26,26 @@ spec:
...
+      cert_manager_test_resources_versioned = cert_manager_test_resources.format(cert_manager_api_version='v1alpha2')
+    else:
+      cert_manager_test_resources_versioned = cert_manager_test_resources.format(cert_manager_api_version='v1')
 
-    # check if cert-mamager is already installed, otherwise pre-load images & apply the manifest
-    # NB. this is required until https://github.com/jetstack/cert-manager/issues/3121 is addressed otherwise
-    # when applying the manifest twice to same cluster kubectl get stuck
-    existsCheck = str(local("kubectl get namespaces", quiet=silent, echo_off=silent))
-    if existsCheck.find("cert-manager") == -1:
```
fabriziopandini, line 36: I'm not sure we can drop this hack for cert manager < v1.0.0
You can test this by doing tilt up twice with cert-manager v0.16.1 (or by applying the cert-manager manifest for v1.16.1 twice on the same cluster).

davideimola: It seems working. As I understand looking at the issue it has been fixed directly on Kubernetes and Helm.
Probably this can create some issues for users with older versions of k8s. I do not know if we want that.

cert_manager/Tiltfile
davideimola: Yeah, I hope someone can suggest something! I have thought on the two following options:

- use the Python semver package if it is possible to import and install it with pip on Tilt
- improve it with regexes

cert_manager/Tiltfile
fabriziopandini: Ok then for me.
It would be nice to add a comment near `kubectl apply -f` to keep track that this could not work when executing "tilt up" more times on the same cluster with older version of kubernetes/certmanager (and link to the issue)

davideimola: Let me know if it is fine.

fabriziopandini (commented):
/lgtm

nicks (approved):
LGTM!

PR #144, opened 2021-03-12
Passthrough allow_duplicates

kevin-lindsay-1:
@nicks 

I just did this through the UI, so I don't know how to also edit the docs in here, but they should be updated, as well.

Whether or not this is a reasonable default is a different question.


nicks (approved):
LGTM!

PR #148, opened 2021-03-16
go and js tests

maiamcc:
docs tk. i'm really curious if these will be useful to people but, they're here.

i don't know a lot about the javascript testing ecosystem--currently i have helpers for running jest tests via npm or yarn, does that seem good enough for a first pass? if not, what other common setups should i cover? (or we should just put this out and see what people ask us for)


tests_in_tilt/Tiltfile
```diff
@@ -0,0 +1,51 @@
+# Some common test configurations to easily get up and running with Tilt for Tests
+# (see docs: https://docs.tilt.dev/tests_in_tilt.html)
+
+### GOLANG TESTS ###
+def test_go_all(path, deps, timeout='', tags=None, mod=''):
```
maiamcc, line 5: Instead of passing `deps` (or if user passes empty `deps`) we could find all `*.go` / `*.js||tsx` files, but testing this in our repo, it takes ~3 seconds to `find` all go files, so probably not worth it?

landism, line 5: 1. The first arg is a package, not a path (e.g., `'./internal/tiltfile'` and `'github.com/tilt-dev/tilt/internal/tiltfile'` are valid, but `'internal/tiltfile'` is not). This is also what go test -help calls its arg. I'm not sure if we're adding much value by forcing it to be recursive if we're already asking the user to know how to specify a go package.
2. nit: I think `timeout` and `mod` should default to `None` or `tags` should default to `[]`, to be consistent.

tests_in_tilt/Tiltfile
```diff
@@ -0,0 +1,51 @@
...
+    if tags:
+        tags_str = '-tags {}'.format(','.join(tags))
+
+    mod_str = ''
+    if mod:
+        mod_str = '-mod {}'.format(mod)
+
+    cmd = 'go test {mod_str} {tags_str} {timeout_str} {path}'.format(
+        mod_str=mod_str, tags_str=tags_str, timeout_str=timeout_str, path=path)
+    test('test-go-all', cmd, deps=deps)
```
landism, line 24: What if I have two sets of tests (e.g. unit and integration)? Maybe parameterize the name or stick the package/path in there?

tests_in_tilt/Tiltfile
```diff
@@ -0,0 +1,51 @@
...
+
+
+### JAVASCRIPT TESTS ###
+def _test_js_all(pkmanager, dir, deps=None, with_install=False):
+    resource_deps = []
+    if with_install:
+        name = pkmanager + '-install'
+        local_resource(name, 'cd {dir} && {pkmanager} install'.format(
```
landism, line 32: we should probably have a dep on yarn.lock / package.json here and then ignore them on the test resource

landism (commented):
seems fine as-is, except waiting for docs to approve

Have you thought much about allowing arbitary arg passthrough? (e.g., `test_go_all('.', '.', extra_args='-v')`)

It'd be pretty annoying to have to explode things if the arg you need doesn't happen to be exposed by the extension. We haven't done this for other things, though. I'm not sure if their reasons translate here.

nicks:
high level concern - we should have separate extensions per language (e.g., test_go, test_js, etc). I think sticking all languages in one extension will scare people off (since I, a JS developer, don't want to accidentally break Ruby)

i also think it's silly to mention "in_tilt" in the name of an extension (not to mention the bad accidental acronym)

tests_in_tilt/Tiltfile
landism: * Even if it were instantaneous, `find -name "*.go"` is kinda annoying in that if I create a new go file after the Tiltfile executes, changes to it won't trigger a build of this resource.
* You could let the user decide by documenting an example that calls `find` or even add a `find` function and have an example call that to set deps.

maiamcc: 1. that makes sense, i changed the arg name and added a `recursive` bool arg in case they want us to add the `/...` for them
2. alas, the [mutable default args](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) gotcha is still a thing in Starlark :( so i think there's a strong argument to be made for "default value of the type, if possible; `None` where it's dangerous". This is the way I've seen it done in lots of other python code fwiw.

tests_in_tilt/Tiltfile
maiamcc: yyyyes but also I have no way of knowing where those files live, since they could be in a parent directory. ugh. let me think.

maiamcc:
@landism does this look right? if yes I'll put in some docs after lunch

tests/javascript/Tiltfile
```diff
@@ -0,0 +1,52 @@
...
+        install_resource = pkmanager + '-install'
+        install_cmd = 'cd {dir} && {pkmanager} install'.format(
+            dir=dir, pkmanager=pkmanager)
+        local_resource(install_resource, install_cmd, install_deps, allow_parallel=True)
+        resource_deps = [install_resource]
+
+        # double check that we could find package.json etc. correctly
+        for f in install_deps:
+            if not os.path.exists(f):
+                warn('''expected JS config file "{}" not found -- did you pass project_root correctly when setting up JS tests?
```
maiamcc, line 18: except it's not really called a "config" file, is it? ugh what's the word?

tests_in_tilt/Tiltfile
landism: > alas, the mutable default args gotcha is still a thing in Starlark :(

computers gonna compute :(

tests/javascript/Tiltfile
landism: This has a nice list of the various files that can be used for this: https://docs.npmjs.com/cli/v7/configuring-npm

But I don't see any sign of a name!

Would it be weird to just say "expected file {} not found"?

maiamcc:
@landism @nicks docs up and ready for review!

landism (commented):
These probably deserve some kind of tests. If setting up the binary deps is bad bang:buck, a tiltfile-result test seems like it'd be pretty cheap?

nicks:
i'll let Matt take the lead on this PR review, but I second this comment that tests would be helpful (both to verify this doesn't break, and as an example to point people at to show how to use it)

tests/README.md
```diff
@@ -0,0 +1,9 @@
+# Tests in Tilt
+
+In this directory, you'll find extensions for some of the most common configurations for running your tests in Tilt. For more about this feature, [see the docs](https://docs.tilt.dev/tests_in_tilt.html).
+
+The extensions here are just starting points for test setup. If you need functionality/test frameworks/support for a language not represented here, let us know.
```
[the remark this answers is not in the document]
maiamcc: oh lol i thought we deleted the contact page for some reason.

tests/Tiltfile
```diff
@@ -0,0 +1 @@
+fail('no top-level Tilt file in this directory--see subdirectories')
```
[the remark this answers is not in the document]
maiamcc: 😅

tests/javascript/Tiltfile
```diff
@@ -0,0 +1,59 @@
...
+        install_cmd = 'cd {dir} && {pkmanager} install'.format(
+            dir=dir, pkmanager=pkmanager)
+        local_resource(install_resource, install_cmd, install_deps, allow_parallel=True)
+        resource_deps = [install_resource]
+
+        # double check that we could find package.json etc. correctly
+        for f in install_deps:
+            if not os.path.exists(f):
+                warn('''expected file "{}" not found -- did you pass project_root correctly when setting up JS tests?
+    If not, we can't watch this file for changes and reinstall JS deps when necessary.'''.format(f))
```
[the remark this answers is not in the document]
maiamcc: > though really I'd probably change the second half to "automatic node module updating disabled" or so. As written it kinda feels like "you screwed up and you need to fix it or else this".
do you think i should DISABLE the updating if i can't find the pkg.json/lockfile? currently all this means is that we can't watch those files to trigger a re-install--but we should be able to run the `yarn|npm install` from anywhere in the project

tests/javascript/Tiltfile
```diff
@@ -0,0 +1,59 @@
...
+def _test_js(name, pkmanager, dir, deps=None, with_install=False, project_root='', ignore=None, extra_args=None):
+    if not project_root:
+        project_root = dir
+
+    install_deps = deps_for_package_manager(pkmanager, project_root)
+    resource_deps = []
+    if with_install:
+        install_resource = pkmanager + '-install'
+        install_cmd = 'cd {dir} && {pkmanager} install'.format(
+            dir=dir, pkmanager=pkmanager)
```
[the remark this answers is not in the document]
maiamcc: mmmm i don't think so, afaik you can call `yarn|npm install` from anywhere within a project, it doesn't have to be the root! (so i guess, see above question about, if we can't determine the project root/package.json etc. should we still run the install, or not?)

tests/javascript/Tiltfile
```diff
@@ -0,0 +1,59 @@
...
+        local_resource(install_resource, install_cmd, install_deps, allow_parallel=True)
```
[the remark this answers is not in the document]
maiamcc: 😬 😬  i do NOT, good catch.

tests/javascript/Tiltfile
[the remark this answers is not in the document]
landism: > currently all this means is that we can't watch those files to trigger a re-install

> we should be able to run the yarn|npm install from anywhere in the project

I'm not sure if I'm misunderstanding the behavior when these files are missing or if I'm misunderstanding the above comment.

My understanding is that if the user specifies the wrong dir, Tilt will not be watching the right paths, and so will not automatically re-run yarn/npm install.

What does "should be able to run the yarn|npm install from anywhere in the project" mean? I'm not sure of an interpretation other than "manually trigger", which wouldn't be an automatic update.

Or are there other deps I'm missing that would trigger automatic updates?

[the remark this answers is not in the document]
maiamcc: > What does "should be able to run the yarn|npm install from anywhere in the project" mean? I'm not sure of an interpretation other than "manually trigger", which wouldn't be an automatic update.

i just mean that we can still do the _installing_, if not the watching+auto-rerun, if the user doesn't give us the correct project_root. e.g. in Tilt if i specify `dir=./web/src`, i won't be able to find `package.json` (it lives in `./web`) but i WILL be able to run `yarn install` from there

so yeah the question right now is, if the user doesn't pass `project_root` and `$DIR/package.json` etc. doesn't exist, do we:
1. run the install cmd (which we can still do) and warn the user that we can't watch `package.json` etc. and so Tilt won't automatically re-run yarn/npm install if deps files change, OR
2. assume that something is pretty wrong and error out
?

thinking about this now, option 2 is maybe safer -- i guess a user can always write their own dependency-installer local resource if their setup is too weird?

tests/javascript/Tiltfile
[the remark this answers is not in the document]
maiamcc: > or it'd probably be cheap and useful to change the build controller so that if A depends on B, A will not start building if B is pending / in progress, but that might be too much of a distraction for this

ugh yes every time i remember that this isn't in-built behavior i cry a little

tests/javascript/Tiltfile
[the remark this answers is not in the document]
landism: Ah ok I meant "automatic updating" as in "automatically run on file changes", but it was ambiguous in that it could also be read as "automatically trigger on startup". Makes sense!

but yeah I feel like option 2 is simplest

tests/javascript/Tiltfile
```diff
@@ -0,0 +1,59 @@
...
+        install_resource = pkmanager + '-install'
+        install_cmd = 'cd {dir} && {pkmanager} install'.format(
+            dir=dir, pkmanager=pkmanager)
+        local_resource(install_resource, install_cmd, install_deps)
+        resource_deps = [install_resource]
+
+        # double check that we could find package.json etc. correctly
+        for f in install_deps:
+            if not os.path.exists(f):
+                fail('''Expected file "{}" not found, refusing to attempt to install JS deps.
```
maiamcc, line 21: @landism better? (i'm a liiiittle worried about what will happen if you're trying to configure a project that doesn't have a lockfile yet, or has a package/lockfile of a different name than we expect...? but we'll cross that bridge when we come to it.)

tests/javascript/Tiltfile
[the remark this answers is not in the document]
maiamcc: cool, done

maiamcc:
@landism if code and docs are okay, can you approve and i'll do tests in a separate PR to keep this one from getting too giant?

tests/javascript/Tiltfile
landism: Ach, sorry, for some reason I was thinking we only did this when project_root was specified (maybe the `project_root = dir` is a bit confusing as a reader?). It seemed like a much clearer case for error when I thought this meant they were specifying a project_root that didn't contain the project files.

Sigh, yeah, a warn is probably best. Sorry for the churn.

tests/javascript/README.md
```diff
@@ -0,0 +1,40 @@
...
+    ```python
+   # 'package.json' and 'yarn.lock' both live in './web/'
+   test_jest_yarn('test-js', './web/src',
+                  with_install=True, project_root='./web')
+    ```
+
+3. Run tests with notifications
+    ```python
+   test_jest_yarn('test-js', './web',
+                  extra_args=['--notify'])
```
landism, line 39: I missed this before, that's neat! I'm half-tempted to write an extension that just wraps local_resource and adds a notification when a build finishes.

landism (approved):

tests/javascript/Tiltfile
maiamcc: 🙃

tests/javascript/README.md
maiamcc: it's jest specific alas

landism: I mean like

```
def notifying_local_resource(name, cmd, **kwargs):
  new_cmd = 'if {}; then terminal-notifier -message "{} succeeded!"; else terminal-notifier -message "{} failed!"; fi'.format(cmd, name, name)
  local_resource(name, new_cmd, **kwargs)
```

PR #152, opened 2021-03-25
tests for javascript test ext (forgive me, father, for I have sinned)

maiamcc:
@landism this is your fault and you should feel bad


landism (approved):
This feels kinda gross at first but I think it's a healthy direction to write more tests in starlark and make it a more compelling option (like, an assert extension with some common assertions)

notice: approval of this PR implies no any acceptance of the labeling of anything as a sin or a father, nor any forgiveness

PR #154, opened 2021-03-30
tests: add run_all param for js tests

maiamcc:
need this so that the CI tests in https://github.com/tilt-dev/tilt-example-nodejs/pull/12 actually do anything useful


nicks:
Can we call this `only_changed`, to match the underlying flag?

https://jestjs.io/docs/cli#--onlychanged

it feels really weird to me to have a thin wrapper around jest that renames the flags to something unrelated to jest.

maiamcc:
Good catch, I had glossed right over the flag name when I first found it. Changed.

nicks (approved):

PR #158, opened 2021-04-13
helm_remote: use watch_settings to ignore the .helm dir

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/watch:

4356b01ddb7552cbdd47b7e0309d4f3024b3168e (2021-04-13 11:46:25 -0400)
helm_remote: use watch_settings to ignore the .helm dir


nicks:
We added watch_settings in September 2020, so I think it's probably safe to add this now?

If we get any complaints about this, we can roll back and do a more explicit version check

nicks:
cc @KOGI

landism (approved):

PR #162, opened 2021-04-23
Allow user to specify which user will be set with restart_process

salim-runsafe:
Adds a `user` argument that will be used to set `USER` in the Dockerfile created with the `restart-helper` files.

I originally opened https://github.com/tilt-dev/tilt/pull/4466 with a different version of this and @maiamcc said I should put a note here to make sure and pull these changes into the integration tests for `tilt-dev/tilt`.

### DIFF
--- a/restart_process/Tiltfile
+++ b/restart_process/Tiltfile
@@ -23,7 +23,7 @@ _CUSTOM_BUILD_KWARGS_BLACKLIST = [
 _ext_dir = os.getcwd()
 
 # shared code between the two restart functions
-def _helper(base_ref, ref, entrypoint, live_update, restart_file=RESTART_FILE, trigger=None, user=root, **kwargs):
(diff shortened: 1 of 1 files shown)


nicks:
Going to assign this to @maiamcc as the owner of this extension!

maiamcc:
@salim-runsafe I'm afraid I sent you on a wild goose chase -- pretty sure this approach won't work because if I create `/.restart-proc` as `root` and then switch back to `USER otheruser` to run my live updates, i'll no longer have permission to write to `/.restart-proc` 🙁 

However, I think #167 solves the problem without adding more layers of user-changing -- what do you think?

salim-runsafe:
Oh yeah, I didn't test that very well @maiamcc. The change in #167 looks like it would work for us. Thanks!

Issue #98, opened 2020-10-30
restart_process: run statements always exec'd as root instead of as USER specified in Dockerfile


maiamcc:
There's already inconsistency in which user Tilt executes `run` statements as (see https://github.com/tilt-dev/tilt/issues/3708): in `--update-mode=container`,  `run` statements are exec'd as `root`; in `--update-mode=exec`, they're exec'd as the container user.

In another wrinkle, when using the `docker/custom_build_with_restart`, Tilt will _always_ exec `run` statements as `root`, because of this change: https://github.com/tilt-dev/tilt-extensions/pull/62 (this change switches the user to `root` before creating the file used to signal process restarts; when trying to create this file as a non-root user (inherited from the parent Dockerfile), Docker would throw a permissions error).

Ideally, `docker/custom_build_with_restart` would leave the container user unchanged from the value specified in the parent Dockerfile (tho given https://github.com/tilt-dev/tilt/issues/3708 that's of limited use :-/ ).

PR #167, opened 2021-04-27
restart_process: solve prob with permissions on restart file by just moving the restart file

maiamcc:
The restart_process extension previously ran into trouble when handling Dockerfiles with non-root users: such a DF wouldn't be able to `touch /.restart-proc` (the default file used to signal a process restart) -- see #52

We worked around this with #62: regardless of the user in the initial Dockerfile, set the user to `root` so we're sure to have permissions for `/.restart-proc`.

This worked okay, but was still a pain for ppl who wanted their subsequent live updates to be run with the user set in their Dockerfile rather than as root (see #98). We even had a user-submitted fix (#162) -- which would cause other, different problems, because presumably live updates as the original non-root user wouldn't have permissions on the `/.restart-proc` file created as `root`.

Rather than increasingly complicated code for changing the Dockerfile user and changing it back, I propose we do the following:
1. change the default path for the file that signals restarts to be `/tmp/.restart-proc` -- `/tmp` is significantly more likely to be writable by a non-root user by default, which obviates our need to change users at all
2. for cases where someone gets permission errors for `/tmp/.restart-proc`, rather than trying to solve the problem by manipulating the dockerfile/the user, instead we direct them to use the `restart_file` param to specify a restart file that they _will_ have permissions for with their current user. (I wish there was a good way to detect this error at runtime and surface useful instructions -- we'll just have to settle for putting it in the README for now :-/ )

TODO: when this is landed, update Tilt integration tests to use the latest version of `restart_process`

### DIFF
--- a/restart_process/README.md
+++ b/restart_process/README.md
@@ -78,7 +78,7 @@ If you see an error of the form:
 ```
 ERROR: Build Failed: ImageBuild: executor failed running [touch /tmp/.restart-proc']: exit code: 1
 ```
-this often means that your Dockerfile user doesn't have permission to write to the file we use to signal a process restart. Use the `restart_file` parameter to specify a file that your Dockerfile user definitely has write access to.
+this often means that your Dockerfile user ([see docs](https://docs.docker.com/engine/reference/builder/#user)) doesn't have permission to write to the file we use to signal a process restart. Use the `restart_file` parameter to specify a file that your Dockerfile user definitely has write access to.
 
 ### API
 ```python
--- a/restart_process/README.md
+++ b/restart_process/README.md
@@ -18,9 +18,8 @@ E.g. if your app is a static binary, you'll probably need to re-execute the bina
 
 ### Unsupported Cases
 This extension does NOT support process restarts for:
-- Images built with `custom_build`
+- Images built with `custom_build` using any of the `skips_local_docker`, `disable_push`, or `tag` parameters.
 - Images run in Docker Compose resources (use the [`restart_container()`](https://docs.tilt.dev/api.html#api.restart_container) builtin instead)
-- `custom_build`
 - Images without a shell (e.g. `scratch`, `distroless`)
 - Container commands specified as `command` in Kubernetes YAML will be overridden by this extension.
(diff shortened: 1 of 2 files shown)


restart_process/README.md
```diff
@@ -73,6 +72,14 @@ custom_build_with_restart(
 )
 ```
 
+### Troubleshooting
+#### `failed running [touch /tmp/.restart-proc']`
+If you see an error of the form:
+```
+ERROR: Build Failed: ImageBuild: executor failed running [touch /tmp/.restart-proc']: exit code: 1
+```
+this often means that your Dockerfile user doesn't have permission to write to the file we use to signal a process restart. Use the `restart_file` parameter to specify a file that your Dockerfile user definitely has write access to.
```
landism, line 81: I worry a bit that "Dockerfile user" isn't really a googleable term. Maybe something like this?

```suggestion
this often means that your Dockerfile user doesn't have permission to write to the file we use to signal a process restart. Use the `restart_file` parameter to specify a file that your Dockerfile user definitely has write access to. You can run `docker run --entrypoint whoami $DOCKER_IMAGE` for your `$DOCKER_IMAGE` to see which user this is, or, e.g., `docker run --entrypoint touch $DOCKER_IMAGE /tmp/.restart-proc` to test permissions.
```

(maybe the last bit is a bit too handholdy?)

landism (approved):
> I wish there was a good way to detect this error at runtime and surface useful instructions

we coooooould replace the touch with a script like this:
```
#!/bin/bash
if [ $# -eq 0 ]; then
  exec touch
fi
touch "$@"
RESULT="$?"
if [ $RESULT -ne 0 ]; then
  # unfortunately it seems that touch just uses exit code 1 for most/all failures
  echo "If this failed with permission denied, see: <insert link>"
  exit $RESULT
fi
```

restart_process/README.md
maiamcc: ehhh kinda handholdy. I just linked to the Dockerfile reference doc for `USER` and if ppl still have questions they can come to us

Issue #169, opened 2021-05-03
restart_process extension and Args


zkutasi:
Hi,

We have the following use-case:
1. We have a helm chart with a deployment, and in this helm chart, we have defined lots of arguments for Java for example
2. We have a Dockerfile, in this dockerfile, we specify the ENTRYPOINT as 
ENTRYPOINT ["/bin/bash", "-e", "./entrypoint.sh"]

3. This entrypoint.sh script is launching our Java app, with arguments all around (JVM plus App parameters)... and we use the fact that by default, Docker execution passes all extra arguments into the entrypoint without specifying it anywhere
4. Now, with Tilt, this does not seem to work, as kind-of stated in the README here: https://github.com/tilt-dev/tilt-extensions/tree/master/restart_process, however you state here that there is a possibility to use Args... but not documented
5. What happens is that in the end, in docker I see this for example with "kubectl describe"
```
 containers:
  - args:
    - -Xms1g -Xmx3g
    - -Dspring.config.location=classpath:application.yaml,file:///etc/spring/custom-application.yaml
    - -Dlog4j.configurationFactory=xxx
    - -Dcore.tls.enabled=false
    - -Dcore.application.name=app
    - -Dcore.logtransformer.port=5014
    - com.company.Application
   command:
    - /tilt-restart-wrapper
    - --watch_file=/tmp/.restart-proc
    - sh
    - -c
    - /bin/bash -e ./entrypoint.sh
   env:

```
and with "docker inspect":

```
        "Path": "/tilt-restart-wrapper",
        "Args": [
            "--watch_file=/tmp/.restart-proc",
            "sh",
            "-c",
            "/bin/bash -e ./entrypoint.sh",
            "-Xms1g -Xmx3g",
            "-Dspring.config.location=classpath:application.yaml,file:///etc/spring/custom-application.yaml",
            "-Dlog4j.configurationFactory=xxx",
            "-Dcore.tls.enabled=false",
            "-Dcore.application.name=app",
            "-Dcore.logtransformer.port=5014",
            "com.company.Application"
        ],

```
As you can see, TIlt replaced the command, and it gets "distributed" into path and args... and the extra arguments coming from helm are "appended" at the end... however, in entrypoint.sh, absolutley no arguments are passed in. I do not really know how these values get merged together into a single command line, how they are escaped, quoted etc...

[...]

PR #172, opened 2021-05-05
restart_process: update readme for troubleshooting args

maiamcc:
documentation to help future users troubleshoot issues like https://github.com/tilt-dev/tilt-extensions/issues/169, i.e.: what to do if my args aren't being correctly applied to my entrypoint command

### DIFF
--- a/restart_process/README.md
+++ b/restart_process/README.md
@@ -80,6 +80,13 @@ ERROR: Build Failed: ImageBuild: executor failed running [touch /tmp/.restart-pr
 ```
 this often means that your Dockerfile user ([see docs](https://docs.docker.com/engine/reference/builder/#user)) doesn't have permission to write to the file we use to signal a process restart. Use the `restart_file` parameter to specify a file that your Dockerfile user definitely has write access to.
 
+#### My args aren't being applied!
+Whatever you pass as the `entrypoint` parameter will override everything in your Dockerfile `COMMAND`, Dockerfile `ARGS`, and Kubernetes YAML `command`.
+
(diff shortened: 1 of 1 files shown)


restart_process/README.md
```diff
@@ -80,6 +80,13 @@ ERROR: Build Failed: ImageBuild: executor failed running [touch /tmp/.restart-pr
 ```
 this often means that your Dockerfile user ([see docs](https://docs.docker.com/engine/reference/builder/#user)) doesn't have permission to write to the file we use to signal a process restart. Use the `restart_file` parameter to specify a file that your Dockerfile user definitely has write access to.
 
+#### My args aren't being applied!
+Whatever you pass as the `entrypoint` parameter will override everything in your Dockerfile `COMMAND`, Dockerfile `ARGS`, and Kubernetes YAML `command`.
+
+If you still want to apply different arguments to different containers, you can specify `args` in your Kubernetes YAML. But _be warned_: these args only work if you pass your entrypoint as args (i.e. as a list of strings). So instead of `entrypoint="/my/great/bin"`*, pass `entrypoint=["/my/great/bin"]`.
```
landism, line 86: This feels like a bug and the docs feel more CYA than anything.

Would one of these work?
1. if entrypoint is a string and contains a space, return an error
2. if entrypoint is a string, `shlex.quote` it

I'd at least like to link to a GH issue here:
1. to make clear that we don't like the behavior, even if we don't have a solution worth implementing right now
2. to allow discussion/rejection of solutions such as above
3. to allow +1s

maiamcc: 1. but you're supposed to be able to pass an entrypoint like `cat myfile.txt > /bin/myapp`, so i don't think we could enforce this (unless at Tiltfile parse, we want to parse the k8s yaml and see if there are `args` specified and only complain then)
2. i'm not sure I understand how this addresses the problem

but yes it's definitely CYA 😅

nicks:
closing this as stale

PR #183, opened 2021-06-18
uibutton: add extension for Cmd buttons [ch12082]

milas:
Extension to make creating custom `UIButton`s for a particular
resource that run a `Cmd` on click easier.

```python
load('ext://uibutton', 'cmd_button')

# define resource 'my-resource'
# k8s_resource('my-resource')

# create a button on resource 'my-resource'
cmd_button(name='Hello World',
           resource='my-resource',
           argv=['echo', 'Hello World'])
```

~~(⚠️ Draft until next Tilt release as this will fail tests since the API isn't fully live yet)~~

### DIFF
--- a/uibutton/Tiltfile
+++ b/uibutton/Tiltfile
@@ -1,0 +1,50 @@
+def cmd_button(name, resource, argv, text=''):
+  if config.tilt_subcommand == 'down':
+    return
+
+  if not text:
+    text = name
+
+  button = {
+    "apiVersion": "tilt.dev/v1alpha1",
+    "kind": "UIButton",
+    "metadata": {
+      "name": name,
+      "annotations": {
+        "tilt.dev/resource": resource,
+      }
+    },
+    "spec": {
+      "text": text,
+      "location": {
+        "componentType": "Resource",
+        "componentID": resource,
+      }
+    }
+  }
+  cmd = {
+    "apiVersion": "tilt.dev/v1alpha1",
+    "kind": "Cmd",
+    "metadata": {
+      "name": "btn-" + name,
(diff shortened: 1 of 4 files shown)


uibutton/Tiltfile
```diff
@@ -0,0 +1,50 @@
...
+    "spec": {
+      "args": argv,
+      "dir": config.main_dir,
+      "startOn": {
+        "startAfter": _now(),
+        "uiButtons": [name],
+      },
+    }
+  }
+  local("echo %s | %s apply -f -" % (shlex.quote(str(encode_yaml_stream([button, cmd]))), sys.executable))
```
nicks, line 44: we're going to need an STDIN stream eventually...

in the meantime, it might be better to do this as an local(env=...) so that the input doesn't show up in the logs...?

nicks (approved):

milas:
Worth noting a weird usability thing is that if you change a button's name _while running `tilt up`_, the old one won't get deleted since the apply is now for a "new" button, so you'll have 2 until you restart.

(Matt's original implementation tried to side-step this by auto-naming buttons with a counter, but Starlark per-context [im]mutability made it impractical when converting to an actual extension.)

uibutton/Tiltfile
milas: Good call - I did `echo_off=True` so the command doesn't show up since it's not particularly useful (_especially_ with the YAML invisible) as well as using `env` to eliminate the need for `shlex.quote` which should reduce the likelihood of a weird edge case escaping YAML 😬 

(Also did `echo_off=True` + `quiet=True` for the `_now()` helper)

PR #185, opened 2021-06-25
ci: update to Go 1.16 and ko 0.8.3

milas:


milas:
This will also bump to latest Tilt which #184 needs 🙃

nicks (approved):

PR #186, opened 2021-06-25
ci: fixes for Go 1.16

milas:
## ci: use pinned versions of images/deps
Prevent surprises with updating CI in the future 🙃 
This also uses a pinned version of the [tilt-dev/tilt-extensions-ci](https://github.com/tilt-dev/tilt-extensions-ci) helper

## ko: fix tests for Go 1.16

The package we were building does not build in Go module-aware
mode. It was also a random third-party package.

This replaces it with a tiny hello world local app. While this
does not test the remote fetch functionality of the extension,
that is actually largely inoperable with modern projects (i.e.
those that use Go modules).

See https://github.com/tilt-dev/tilt-extensions/issues/188 for details.

## test-in-tilt: fix for Go 1.16

These were running module-mode incompliant commands - when you
have a `go.mod`, you MUST run the `go test` command from within
the module hierarchy (i.e. at or under the `go.mod` dir).

In practice, this isn't a huge issue, as generally you'd have a
`Tiltfile` at the same level as a `go.mod`, so the test structure
is adjusted to reflect this with the elimination of the `fake_project`
dir.

It is possible to work around this using the new `dir` argument
that `local_resource` supports, which will eventually receive it
via `**kwargs`, so if you have the file structure:

```
├── go-app
│   ├── go.mod
│   └──main.go
└── Tiltfile
```

You can do:
```
test_go('mytests', './...', './go-app', dir='./go-app')
```

So that the `go test ./...` command is run inside of the `go-app`
directory, instead of trying to pass `./go-app/...` as the "package".


nicks (approved):

.circleci/config.yml
```diff
@@ -3,12 +3,12 @@ jobs:
   check:
     docker:
       - image: tiltdev/tilt-extensions-ci:latest
-        
+
     steps:
       - checkout
-      - run: GOBIN=$HOME/.bin go get github.com/tilt-dev/tilt-extensions-ci && go install github.com/tilt-dev/tilt-extensions-ci
+      - run: GOBIN=$HOME/.bin go get github.com/tilt-dev/tilt-extensions-ci && go install 'github.com/tilt-dev/tilt-extensions-ci@latest'
```
nicks, line 9: we should probably change this to use a SHA hash like we do in the main tilt repo -- it makes it a bit easier to make changes atomically

milas: Using Docker SHA + Git tag (for `tilt-extensions-ci`) w/ 9029b1d

milas:
@nicks Okay, I had to make some other changes for `ko` + `tests-in-tilt/go` due to module incompliant stuff they were both doing that doesn't fly in 1.16+. Updated PR description has all the details.

This should make Go 1.17 smoother at least!

PR #189, opened 2021-06-28
coreos_prometheus: cross-platform support + misc fixes

### LINKED ISSUE #130: Prometheus extension example fails on macOS
output:
```
...
local: tar -C /var/folders/f8/pqx_d6kx5hgdvghjhhf0qd8h0000gn/T/tmp.lBoYMb6e/prometheus-operator-kube-prometheus-v0.5.0 -xzf /Users/user/code/repos/flare/tilt_modules/coreos_prometheus/.tilt-cache/coreos-kube-prometheus-v0.5.0.tar.gz --strip-components=1 --wildcards 'prometheus-operator-kube-prometheus-*/manifests'
 → tar: Option --wildcards is not supported
 → Usage:
 →   List:    tar -tf <archive-filename>
 →   Extract: tar -xf <archive-filename>
 →   Create:  tar -cf <archive-filename> [filenames...]
 →   Help:    tar --help
Successfully loaded Tiltfile (4.120123106s)
  /Users/user/code/repos/flare/Tiltfile:15:1: in <toplevel>
  /Users/user/code/repos/flare/tilt_modules/coreos_prometheus/Tiltfile:176:25: in <toplevel>
  /Users/user/code/repos/flare/tilt_modules/coreos_prometheus/Tiltfile:66:10: in download_files
  <builtin>: in local
Error: command ["sh" "-c" "tar -C /var/folders/f8/pqx_d6kx5hgdvghjhhf0qd8h0000gn/T/tmp.lBoYMb6e/prometheus-operator-kube-prometheus-v0.5.0 -xzf /Users/user/code/repos/flare/tilt_modules/coreos_prometheus/.tilt-cache/coreos-kube-prometheus-v0.5.0.tar.gz --strip-components=1 --wildcards 'prometheus-operator-kube-prometheus-*/manifests'"] failed.
error: exit status 1
```

I haven't traced down the source of this `tar` usage yet to look at a patch but the issue looks as simple as usage of the non-portable `--wildcards` flag which won't work on macOS; perhaps an alternative syntax could be used here (but not sure what other issues might follow).

milas:
* Fix logic to find root Tiltfile (it would just recurse up to
  filesystem root)
* Ensure we always try to create cache dir
* Use GitHub release download URL (this is faster and helps solve
  issues with tarfile structure)
* Use a cross-platform tar command (possible since the pattern
  is now predictable as it doesn't have a Git commit hash but the
  tag name instead)
* Suppress most command echo for init process
* Only output on error for download via curl
* Run test as part of CI

Fixes #130.

### DIFF
--- a/coreos_prometheus/Tiltfile
+++ b/coreos_prometheus/Tiltfile
@@ -4,7 +4,7 @@ def _find_root_tiltfile_dir():
     # Find top-level Tilt path
     current = os.path.abspath('./')
     while current != '/':
-        if os.path.exists(os.path.join(current, '')):
+        if os.path.exists(os.path.join(current, 'Tiltfile')):
             return current
 
         current = os.path.dirname(current)
@@ -12,16 +12,15 @@ def _find_root_tiltfile_dir():
     fail('Could not find root Tiltfile')
 
 def _find_cache_dir():
-    from_env = os.getenv('TILT_CACHE_DIR', '')
+    cachedir = os.getenv('TILT_CACHE_DIR', '')
 
-    if from_env != '':
-        return from_env
+    if cachedir == '':
+        cachedir = os.path.join(_find_root_tiltfile_dir(), '.tilt-cache')
 
-    cachedir = os.path.join(_find_root_tiltfile_dir(), '.tilt-cache')
     if not os.path.exists(cachedir):
-        local('mkdir -p %s' % cachedir, quiet=True)
-    os.putenv('TILT_CACHE_DIR', cachedir)
+        local('mkdir -p %s' % cachedir, echo_off=True)
 
+    os.putenv('TILT_CACHE_DIR', cachedir)
     return cachedir
 
 def _create_temp_dir():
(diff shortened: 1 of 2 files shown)


milas:
I noticed this was missing from the CI run last week (was comparing the list after realizing I'd forgotten to include `uibutton` in there and then with all the Go 1.16 stuff) and so took this as an opportunity to get it working on macOS so it was easier for me to test as well 😉

nicks (approved):

milas:
@nicks Apparently this exceeds limits in our CI - maybe that's why we never had Prometheus tests on before?

```
prometheus │ [K8s EVENT: Pod prometheus-k8s-1 (ns: monitoring)] 0/1 nodes are available: 1 Insufficient cpu.
```

From what I can tell, KIND will infer from the host and the [`setup_remote_docker` host](https://circleci.com/docs/2.0/building-docker-images/#specifications) is always 2CPU/8GB RAM regardless of executor choice.

Thoughts on continuing to exclude it from CI for now? Or is there a trick to overcommit/bypass limits on KIND that you know of?

nicks:
let's take it off CI for now. we can probably override the cpu requirements in the test yaml

PR #190, opened 2021-06-28
configmap: Add envfile support

BitFlippedCelt:
I ran into a local development use-case where being able to leverage kubectl from-env-file support within config map generation became very relevant. I have made the updates and extended docs and testing suite to support this updated functionality. The main gotcha with the env file loader is that it supports a singular string instead of a list of strings. This is primarily to mitigate confusion with kubectl's behavior of only loading the last provided env file when it has been provided a list of files.

### DIFF
--- a/configmap/README.md
+++ b/configmap/README.md
@@ -9,18 +9,19 @@ Helper functions for creating Kubernetes configmaps.
 ### configmap_yaml
 
 ```
-configmap_yaml(name: str, namespace: str = "", from_file: Union[str, List[str]] = None, watch: bool = True): Blob
+configmap_yaml(name: str, namespace: str = "", from_file: Union[str, List[str]] = None, from_env_file: str = None, watch: bool = True): Blob
 ```
 
 Returns YAML for a config map generated from a file.
 
 * `from_file` ( str ) – equivalent to `kubectl create configmap --from-file`
+* `from_env_file` (str) - equivalent to `kubectl create configmap --from-env-file`
 * `watch` ( bool ) - auto-reload if the files change
 
 ### configmap_create
 
 ```
-configmap_create(name: str, namespace: str = "", from_file: Union[str, List[str]] = None, watch: bool = True)
+configmap_create(name: str, namespace: str = "", from_file: Union[str, List[str]] = None, from_env_file: str = None, watch: bool = True)
 ```
 
 Deploys a config map. Equivalent to
--- a/configmap/README.md
+++ b/configmap/README.md
@@ -9,7 +9,7 @@ Helper functions for creating Kubernetes configmaps.
 ### configmap_yaml
 
 ```
-configmap_yaml(name: str, namespace: str = "", from_file: Union[str, List[str]] = None, from_env_file: str = None, watch: bool = True): Blob
+configmap_yaml(name: str, namespace: str = "", from_file: Union[str, List[str]] = None, watch: bool = True, from_env_file: str = None): Blob
 ```
 
 Returns YAML for a config map generated from a file.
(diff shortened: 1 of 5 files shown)


configmap/Tiltfile
```diff
@@ -1,6 +1,6 @@
 # -*- mode: Python -*-
 
-def configmap_yaml(name, namespace="", from_file=None, watch=True):
+def configmap_yaml(name, namespace="", from_file=None, from_env_file=None, watch=True):
```
milas, line 3: To prevent issues with anyone using positional args (e.g. `configmap_yaml('foo', 'default', './myconfig', True)`), can you change `from_env_file` to be the final argument?

configmap/Tiltfile
```diff
@@ -39,22 +43,36 @@ def configmap_yaml(name, namespace="", from_file=None, watch=True):
...
+      generator = True
 
   if not generator:
     fail("No configmap generator specified")
 
   args.extend(["-o=yaml", "--dry-run=client"])
   return local(args, quiet=True)
 
-def configmap_create(name, namespace="", from_file=None, watch=True):
+def configmap_create(name, namespace="", from_file=None, from_env_file=None, watch=True):
```
milas, line 61: Same thing here - `from_env_file` will need to be last to keep compatibility for positional arguments.

milas (changes requested):
Thanks for your contribution - this is a nice addition! 🙂

configmap/Tiltfile
BitFlippedCelt: I'm made the requested update, good catch! Was not thinking about the positional implication here!

configmap/README.md
```diff
@@ -9,18 +9,19 @@ Helper functions for creating Kubernetes configmaps.
 ### configmap_yaml
 
 ```
-configmap_yaml(name: str, namespace: str = "", from_file: Union[str, List[str]] = None, watch: bool = True): Blob
+configmap_yaml(name: str, namespace: str = "", from_file: Union[str, List[str]] = None, from_env_file: str = None, watch: bool = True): Blob
```
milas, line 12:
```suggestion
configmap_yaml(name: str, namespace: str = "", from_file: Union[str, List[str]] = None, watch: bool = True, from_env_file: str = None): Blob
```

configmap/README.md
```diff
@@ -9,18 +9,19 @@ Helper functions for creating Kubernetes configmaps.
...
 
 * `from_file` ( str ) – equivalent to `kubectl create configmap --from-file`
+* `from_env_file` (str) - equivalent to `kubectl create configmap --from-env-file`
 * `watch` ( bool ) - auto-reload if the files change
 
 ### configmap_create
 
 ```
-configmap_create(name: str, namespace: str = "", from_file: Union[str, List[str]] = None, watch: bool = True)
+configmap_create(name: str, namespace: str = "", from_file: Union[str, List[str]] = None, from_env_file: str = None, watch: bool = True)
```
milas, line 24:
```suggestion
configmap_create(name: str, namespace: str = "", from_file: Union[str, List[str]] = None, watch: bool = True, from_env_file: str = None)
```

milas (commented):
(Updated the docs to keep the argument order in sync with source)

BitFlippedCelt:
Hold up on this, investigating a failure state during `tilt down`.

BitFlippedCelt:
Ok I had missed the positional placement update within the `configmap_create` which in turn set the `from_env_file` value to true when using named variables within the test suite. Should be ok now!

milas (approved):

milas:
Merged! Thanks again for your contribution 👍

PR #192, opened 2021-06-29
ci: add timeout to test execution

milas:
After a bad change the other day, CI got stuck running tests for
~3 hours before I noticed.

CircleCI doesn't have a generic timeout option in YAML config,
they recommend using the `timeout` command if not using a test
framework that already supports a specific option.

As of June 2021, tests take ~7m to run in CI, so this gives a
big buffer by capping at 15m at which point it'll send `SIGTERM`.
If the tests don't shut down cleanly after another minute, it'll
send a `SIGKILL`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
       - setup_remote_docker:
           version: '20.10.6'
       - checkout
-      - run: ctlptl create cluster kind --registry=ctlptl-registry && ./test.sh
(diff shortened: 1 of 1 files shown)


nicks (approved):
LGTM, but maybe we should eat our own dogfood and set a tilt ci default timeout as in https://github.com/tilt-dev/tilt/issues/3783

nicks:
@milas were you planning to submit this?

milas:
Ah, yeah, I'll go ahead and merge this for now to prevent runaway CI, and we can transition to use `tilt ci` arg if we add that at some point

PR #194, opened 2021-06-30
Revert "tilt_inspector: update"

nicks:
Reverts tilt-dev/tilt-extensions#191


nicks:
not totally sure i understand what's going on but will revert for now

PR #198, opened 2021-07-06
add port forwards listing

lucieleblanc:
Creates a new resource that lists all port forwards in use.

### DIFF
--- a/list_port_forwards/display_port_forwards.sh
+++ b/list_port_forwards/display_port_forwards.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+echo "__(.)= __(.)=    =(.)__ =(.)__"
+echo "\___)  \___)      (___/  (___/"
+
+echo "---- Port forwards in use ----"
+
+jqquery='.items[] | { name:.metadata.annotations."tilt.dev/resource", p:.spec.forwards[] } | "\(.name):\(.p.containerPort) -> \(.p.host):\(.p.localPort)"'
+
+startquote='s/^"//'
+endquote='s/"$//'
+
+tilt get portforwards -o json | jq "$jqquery" | sed -e "$startquote" -e "$endquote"
+
+echo "------------------------------"
--- a/list_port_forwards/display_port_forwards.sh
+++ b/list_port_forwards/display_port_forwards.sh
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
+
+command -v jq >/dev/null || (echo "jq not installed :(" && exit 1)  # ensure jq is installed
+
 echo "__(.)= __(.)=    =(.)__ =(.)__"
 echo "\___)  \___)      (___/  (___/"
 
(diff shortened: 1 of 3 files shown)


list_port_forwards/display_port_forwards.sh
```diff
@@ -0,0 +1,15 @@
...
+echo "\___)  \___)      (___/  (___/"
+
+echo "---- Port forwards in use ----"
+
+jqquery='.items[] | { name:.metadata.annotations."tilt.dev/resource", p:.spec.forwards[] } | "\(.name):\(.p.containerPort) -> \(.p.host):\(.p.localPort)"'
+
+startquote='s/^"//'
+endquote='s/"$//'
+
+tilt get portforwards -o json | jq "$jqquery" | sed -e "$startquote" -e "$endquote"
```
KOGI, line 13: This assumes the presnce of `jq` on the user's machine...

list_port_forwards/display_port_forwards.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
nicks, line 1: bash has tricky error-handling. here's a good explanation:
https://wizardzines.com/comics/bash-errors/

I'd probably add
set -euo pipefail
here, which is closer to the semantics you want

list_port_forwards/display_port_forwards.sh
```diff
@@ -0,0 +1,15 @@
...
+
+echo "__(.)= __(.)=    =(.)__ =(.)__"
+echo "\___)  \___)      (___/  (___/"
+
+echo "---- Port forwards in use ----"
+
+jqquery='.items[] | { name:.metadata.annotations."tilt.dev/resource", p:.spec.forwards[] } | "\(.name):\(.p.containerPort) -> \(.p.host):\(.p.localPort)"'
+
+startquote='s/^"//'
```
nicks, line 10: i think `jq -r` will omit the quotes for you, so you don't need to do it manually

list_port_forwards/display_port_forwards.sh
nicks: ya, maybe add a README that mentions that this extension depends on `jq`?

also, add a check at the top like

```
command -v jq || (echo "jq not installed :(" && exit 1)
```

nicks (commented):
ooh this is a cool demo on how to use the portforward api

/cc @ellenkorbes might have other ideas on how the dev interaction should work

ellenkorbes:
Question: Can you have a UIButton under the Tiltfile resource? IMHO that'd be the most natural place for this to live.

But that's just my opinion. This is fine as is.

list_port_forwards/display_port_forwards.sh
lucieleblanc: Thank you for the explanation!

list_port_forwards/display_port_forwards.sh
lucieleblanc: Ah, thank you for pointing that out -- I've added the check.

list_port_forwards/display_port_forwards.sh
lucieleblanc: Changed. Thank you!

nicks (approved):

PR #199, opened 2021-07-07
fix README.md example

justinschuldt:
Thanks for this!


milas (approved):
Whoops, good catch! Thanks for the PR 🙂

PR #203, opened 2021-07-14
dependencies: button to list resource dependencies

lucieleblanc:
Adds a UIButton to print which resources are waiting for a dependency.


list_dependencies/README.md
```diff
@@ -0,0 +1,17 @@
...
+1. find each resource's dependencies
+```
+$ tilt dump engine | jq '.ManifestTargets[].Manifest | { name:.Name , deps:.ResourceDependencies } '
+```
+2. find each resource's status
+```
+$ tilt dump engine | jq '.ManifestTargets[] | { resource_name:.Manifest.Name, pods:.State.RuntimeState.Pods } | .pods["\(.resource_name)"].containers[].ready'
+```
+
+This extension requires `python3` and its `subprocess` and `json` libraries.
```
KOGI, line 17: It might be nice to provide a sample output here, to help potential users decide if this does what they want

lucieleblanc: Just added sample output. Thank you!

list_dependencies/deps.py
```diff
@@ -0,0 +1,60 @@
...
+
+    for resInfo in data.values():
+        manifest = resInfo["Manifest"]
+        resName = manifest["Name"]
+        resDeps = manifest["ResourceDependencies"]
+        if resDeps == None:
+            resDeps = []
+        dependencies[resName] = resDeps
+
+        pods = resInfo["State"]["RuntimeState"]["Pods"]
```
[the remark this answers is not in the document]
nicks: similarly, i wonder if we should put ResourceDependencies on the UIResourceStatus type, so you can also read those with `tilt get`. wdyt? 

i think you can only get ResourceDependencies from 'dump engine' right now. but we're generally moving away from 'tilt dump engine' - the problem is that it's one big blob of data, whereas we're moving towards smaller API objects that you can listen to updates on without caring about the others.

list_dependencies/test/Tiltfile
```diff
@@ -0,0 +1,21 @@
+load('../Tiltfile', 'display_dependencies')
```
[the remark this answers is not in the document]
lucieleblanc: Ah, I didn't realize all the others had `test.sh` in their `test/` directory. I originally intended this as an example for manual testing, but I've added a test script.

list_dependencies/deps.py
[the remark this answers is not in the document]
lucieleblanc: > can this use 'tilt get kubernetesdiscovery' to fetch the pods? we're much more likely to support that long-term...

Changed. Pod status is now obtained using a separate call to `tilt get kubenetesdiscovery -o json`.

> i wonder if we should put ResourceDependencies on the UIResourceStatus type, so you can also read those with tilt get. wdyt? [...] i think you can only get ResourceDependencies from 'dump engine' right now.

Yes, as far as I can tell, none of the `tilt get` commands' objects currently include dependency information, which is why I used `tilt dump engine` here instead. It would be nice to have dependencies in a new or existing API object.

list_dependencies/deps.py
```diff
@@ -0,0 +1,84 @@
...
+
+    for resInfo in ready_data:
+        resName = resInfo["metadata"]["name"]
+        pods = resInfo["status"]["pods"]
+        resStatus = False
+        if pods != None:
+            for pod in pods:
+                containers = pod["containers"]
+                for container in containers:
+                    resStatus = container["ready"]
```
lucieleblanc, line 47: @nicks is it possible to have multiple pods for a resource? If so, do all pods for a resource need to be ready before dependents are allowed to build? Maybe I should change this line to `resStatus &= container["ready"]` to make sure the status reflects the entire resource.

list_dependencies/test/test.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
```
nicks, line 1: you should also add this to the root test.sh ! 
https://github.com/tilt-dev/tilt-extensions/blob/master/test.sh

list_dependencies/deps.py
nicks: for this demo, it's ok to assume that if one pod is ready, then anything that depends on it can start building.

the real relationship between readiness and resource dependencies is a bit slippery. pods can go from crashing to ready back to crashing. whereas once the resource starts building, we don't try to claw it back if the stuff it depends on starts crashing.

list_dependencies/deps.py
```diff
@@ -0,0 +1,84 @@
...
+            stdout=subprocess.PIPE,
+        ).stdout.decode('utf-8')
+    node_data = json.loads(node_str)["ManifestTargets"]
+
+    dependencies = {}
+
+    for resInfo in node_data.values():
+        manifest = resInfo["Manifest"]
+        resName = manifest["Name"]
+        resDeps = manifest["ResourceDependencies"]
```
nicks, line 23: would you like to add ResourceDependencies to UIResource :smile: 

but ya, this is fine for now

nicks (approved):

list_dependencies/deps.py
lucieleblanc: I'll add a note about this in the README. Thank you!

list_dependencies/test/test.sh
lucieleblanc: Done, thank you!

list_dependencies/deps.py
lucieleblanc: Yes, I'll add it and start a new PR here once that is done.

PR #207, opened 2021-07-22
workflow: buttons for a simple linear workflow

lucieleblanc:
Bugs/limitations:
- can't `replay` immediately after running the `reset` command


workflow/README.md
```diff
@@ -0,0 +1,20 @@
+## Workflow
+
+Attach a workflow, or series of commands, to a resource.
+
+```
+workflow(workflow_name : string, resource_name : string, work_cmds=[[]] : [[string]], clear_cmd=[]: [string])
```
milas, line 6: It might be helpful to include an example invocation here as guidance on the types, e.g. that `work_cmds` is an array of argv-style arrays

workflow/Tiltfile
```diff
@@ -0,0 +1,79 @@
+load('ext://uibutton', 'cmd_button')
+
+def set_work_cmds(flow_name, res_name, work_cmds):
+    cmds = ['./write_workflow.sh', flow_name, res_name]
+    for i in work_cmds:
+        cmds.append(" ".join(i))
```
milas, line 6: I think you might need to do something like `cmds.append(" ".join(shlex.quote(arg) for arg in i))` to prevent escaping issues.

(Try having a work_cmd that's something like `['/bin/sh', '-c', 'echo "Hello World"']` for example - I _think_ that might be sufficient to break it)

workflow/Tiltfile
```diff
@@ -0,0 +1,79 @@
...
+    local(cmds)
```
milas, line 7: It might be nice to pass `echo_off=True` to any of these "internal" type commands to reduce output in logs

workflow/play.sh
```diff
@@ -0,0 +1,12 @@
...
+
+set -eou pipefail
+
+if [ $# -ne 2 ] 
+then
+	echo "usage: $0 <workflow_name> <resource_name>"
+	exit 1
+fi
+
+./increment_index.sh $@
```
milas, line 11: I'd rebase on/merge `master` - it runs `shellcheck` on things now via CI (and has `make shellcheck` for local). 

(I'm 99% sure it's going to want this quoted)

workflow/README.md
lucieleblanc: Added. Thank you!

workflow/exec_workflow.sh
```diff
@@ -14,7 +14,6 @@ workflow_name=$2
 indexfile="workflow_index-$resource_name-$workflow_name.tmp"
 cmdsfile="workflow_cmds-$resource_name-$workflow_name.tmp"
 
-WORKFLOW_INDEX=$(cat $indexfile | head -n 1)
-# echo "executing command $WORKFLOW_INDEX: '$( head -n $WORKFLOW_INDEX $cmdsfile | tail -n 1 )'"
-$(head -n $WORKFLOW_INDEX $cmdsfile | tail -n 1)
+WORKFLOW_INDEX=$(head -n 1 <"$indexfile")
```
milas, line 17: I noticed if I rewind then replay immediately, I get an error (since index is `0`):
```sh
# reset
Running cmd: /Users/milas/dev/tilt-extensions/workflow/rewind.sh test (Tiltfile)
resetting...

# immediate replay
Running cmd: /Users/milas/dev/tilt-extensions/workflow/replay.sh test (Tiltfile)
head: illegal line count -- 0
/Users/milas/dev/tilt-extensions/workflow/replay.sh test (Tiltfile) exited with exit code 1
```

I think maybe just exiting with a "no step to replay" type message if `$indexfile` is `0` would be sufficient.

milas (approved):

workflow/exec_workflow.sh
lucieleblanc: Done. Thank you!

PR #208, opened 2021-07-23
Allow configmap extension to use from-literal

davidamin:
Hi all!

We're big fans of this extension at my company. Internally, we've found it handy to be able to leverage the `--from-literal` option on the same command. This change adds a new function to make that option accessible through the module. 

Thanks!


landism (approved):
LGTM, thanks!

PR #206, opened 2021-07-21
add shellcheck

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/shellcheck:

018a67d9a544f9f1df6238074db9e781ef7af60b (2021-07-21 14:15:58 -0400)
add shellcheck

Code review reminders, by giving a LGTM you attest that:

* Commits are adequately tested
* Code is easy to understand and conforms to style guides
* Incomplete code is marked with TODOs
* Code is suitably instrumented with logging and metrics

PR #210, opened 2021-07-26
syncback: fix namespace escaping

### LINKED ISSUE #209: Syncback - Issue copying rsync binary into container after recent changes
I was working through Syncback errors with a coworker today and she noticed that this recent commit quotes some arguments in a way that breaks correct argument parsing. After reverting krsync.sh to the version int he last commit, it worked for me again.

You can see here that the last part is quoted, but should not be since it is not actually a single argument but many.
https://github.com/tilt-dev/tilt-extensions/blame/018a67d9a544f9f1df6238074db9e781ef7af60b/syncback/krsync.sh#L18

landism:
Fixes #209

### Problem

The [shellcheck PR](#206) [quoted `$namespace` in syncback's krsync.sh](https://github.com/tilt-dev/tilt-extensions/commit/018a67d9a544f9f1df6238074db9e781ef7af60b#diff-1735f7dcd2160cebc05a56e9882d2092d4725f9b8104ea5a22e7b81e7fb85864L43). It turns out `$namespace` is not a namespace but either `''` or `'-n some_namespace'`. Quoting the latter breaks the script, yielding this error:
```
error: pod, type/name or --filename must be specified
```

### Solution

Use the syntax described in the [exceptions section for SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086#exceptions).

I skimmed the [shellcheck PR](#206) and didn't see any other instances of this issue.

### DIFF
--- a/syncback/krsync.sh
+++ b/syncback/krsync.sh
@@ -19,7 +19,7 @@ if [ -z "$KRSYNC_STARTED" ]; then
 fi
 
 # Running as --rsh
-namespace=''
+namespace_args=()
 depl=$1
 shift
 
@@ -27,7 +27,7 @@ shift
 if [ "X$depl" = "X-l" ]; then
     depl=$1
     shift
-    namespace="-n $1"
+    namespace_args=(-n "$1")
     shift
 fi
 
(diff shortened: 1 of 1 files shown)


nicks (approved):

nicks:
will investigate what it would take to test for this

PR #213, opened 2021-08-02
workflow: fix shellcheck violation

milas:
Use `eval` on the command that is grabbed from the list of commands.

NOTE: The current approach actually has issues with multiline commands;
this is neither fixed nor exacerbated by this change - this is purely
to make shellcheck happy.

### DIFF
--- a/workflow/exec_workflow.sh
+++ b/workflow/exec_workflow.sh
@@ -21,5 +21,5 @@ if [[ "$WORKFLOW_INDEX" -eq "0" ]] ; then
 	exit 0
 fi
 
-$(head -n "$WORKFLOW_INDEX" "$cmdsfile" | tail -n 1)
-
+cmd=$(head -n "${WORKFLOW_INDEX}" "${cmdsfile}" | tail -n 1)
+eval "${cmd}"


workflow/exec_workflow.sh
```diff
@@ -21,5 +21,5 @@ if [[ "$WORKFLOW_INDEX" -eq "0" ]] ; then
 	exit 0
 fi
 
-$(head -n "$WORKFLOW_INDEX" "$cmdsfile" | tail -n 1)
-
+cmd=$(head -n "${WORKFLOW_INDEX}" "${cmdsfile}" | tail -n 1)
```
nicks, line 24: hmmm...why do we need to subshell this at all?

nicks (approved):

workflow/exec_workflow.sh
milas: The extension creates a file with all the commands from the Tiltfile and then grabs the command to exec by index. 

cmdsfile:
```
echo "step 1"
echo "step 2"
```
so `cmd=$(head -n 2 cmdsfile | tail -n 1)` results in `cmd='echo "step 2"'` and then that gets eval'd.

(See also PR note: this actually causes trouble with multiline scripts - I think it probably needs to change to create temporary `stepN.sh` scripts and run those directly)

PR #217, opened 2021-08-04
Accept all local args in local_output

davidamin:
This extension is super handy for grabbing the output of commands, but it doesn't have argument parity with the builtin `local`. It's often super handy to be able to pass in a new env var or set `quiet`, this PR adds the same default args as `local` and passes them through


milas (approved):
Thanks for the PR!

PR #218, opened 2021-08-05
Provide ability to pass custom entr args to restart wrapper.

mhaley-tignis:
This PR provides the ability for the user to supply custom arguments to entr when needed.

This is needed for some of our python development workflows that rely on `pip install -e .`, and only need the live reload to copy the new app files into the container, then restart the main entrypoint. The problem we encounter is that if the application crashes, entr exits, tilt loses its active container and will rebuild everything from scratch, which takes far too long. If I pass in just `-r` to entr, when the app under development crashes, `entr` keeps running, leaving the container up which lets us continue our live reload loop instead of having to rebuild everything.

### DIFF
--- a/restart_process/Tiltfile
+++ b/restart_process/Tiltfile
@@ -29,7 +29,8 @@ def _helper(base_ref, ref, entrypoint, live_update, restart_file=RESTART_FILE, t
     # declare a new docker build that adds a static binary of tilt-restart-wrapper
     # (which makes use of `entr` to watch files and restart processes) to the user's image
     df = '''
-    FROM tiltdev/restart-helper:2021-08-05v2 as restart-helper
+    FROM tiltdev/restart-helper:2021-08-09 as restart-helper
+
     FROM {}
     RUN ["touch", "{}"]
     COPY --from=restart-helper /tilt-restart-wrapper /
@@ -67,6 +68,7 @@ def docker_build_with_restart(ref, context, entrypoint, live_update,
                               trigger=None, entr_flags=None, **kwargs):
     """Wrap a docker_build call and its associated live_update steps so that the last step
     of any live update is to rerun the given entrypoint.
+
      Args:
       ref: name for this image (e.g. 'myproj/backend' or 'myregistry/myproj/backend'); as the parameter of the same name in docker_build
       context: path to use as the Docker build context; as the parameter of the same name in docker_build
--- a/restart_process/Tiltfile
+++ b/restart_process/Tiltfile
@@ -22,7 +22,7 @@ _CUSTOM_BUILD_KWARGS_BLACKLIST = [
 _ext_dir = os.getcwd()
 
 # shared code between the two restart functions
-def _helper(base_ref, ref, entrypoint, live_update, restart_file=RESTART_FILE, trigger=None, entr_flags=None, **kwargs):
+def _helper(base_ref, ref, entrypoint, live_update, restart_file=RESTART_FILE, trigger=None, exit_policy='restart', **kwargs):
     if not trigger:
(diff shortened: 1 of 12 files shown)


restart_process/Tiltfile
```diff
@@ -22,15 +22,14 @@ _CUSTOM_BUILD_KWARGS_BLACKLIST = [
...
+def _helper(base_ref, ref, entrypoint, live_update, restart_file=RESTART_FILE, trigger=None, entr_flags=None, **kwargs):
     if not trigger:
         trigger = []
 
     # declare a new docker build that adds a static binary of tilt-restart-wrapper
     # (which makes use of `entr` to watch files and restart processes) to the user's image
     df = '''
-    FROM tiltdev/restart-helper:2020-10-16 as restart-helper
-
+    FROM tiltdev/restart-helper:2021-08-05v2 as restart-helper
```
mhaley-tignis, line 32: We will need to push a new version of the restart helper, then once it is up I can update this to the right value.

mhaley-tignis: This is also why the tests are failing
```  failing-job │ ERROR: Build Failed: ImageBuild: failed to create LLB definition: docker.io/tiltdev/restart-helper:2021-08-05v2: not found```

File: restart_process/Tiltfile [hunk 2]
```diff
@@ -62,10 +64,9 @@ def _helper(base_ref, ref, entrypoint, live_update, restart_file=RESTART_FILE, t
 
 def docker_build_with_restart(ref, context, entrypoint, live_update,
                               base_suffix='-tilt_docker_build_with_restart_base', restart_file=RESTART_FILE,
-                              trigger=None, **kwargs):
+                              trigger=None, entr_flags=None, **kwargs):
```
nicks, line 67: Thanks for the PR @mhaley-tignis !

I'm a little bit concerned about whether people will be able to use this API successfully. You really need to deeply understand both how entr works and how your underlying live-reloader works

I remember it being a deep discussion when we wrote this extension. AFAIR, some live-reload entrypoints only exit on catastrophic error (i.e., live-reload isn't possible at all).

@landism do you remember the details of this discussion?

mhaley-tignis: `-z` will be there so that the container exits when the process finishes. This is useful for job containers, or init containers that you expect to exit when done.

nicks: i wonder if it makes sense for the api here to be something like `restart_policy="always"` or `restart_on_exit=True`?

in any case, lemme take care of the container image separately

landism: > @landism do you remember the details of this discussion?

sorry, doesn't ring a bell and I don't see any mention of it in the doc that was written at the time

hiding entr flags from the user sounds good, though

mhaley-tignis: I like the idea of using `restart_policy="always"` because it is more flexible for future expansion if it is ever needed, but I think that it is confusing from a user perspective. As a user I could get confused and think the `restart_policy` applies to _when_ tilt decides to restart the entrypoint, same with `restart_on_exit`. We really aren't restarting anything, maybe we want something like `proc_exit_policy`?

nicks: how about `exit_policy="restart"` for the new behavior? and can you add this to the custom_build_with_restart function while you're here?

nicks:
if you rebase on the code i just merged, you should be able to make this change Tiltfile-only!

mhaley-tignis:
Thank you for the support @nicks. Tests pass and changes are ready for review once more.

nicks (approved):
LGTM!

PR #222, opened 2021-08-10
local_output: fix empty serve_cmd

landism:
### Problem

#217 defaulted command_bat to the empty string, which can cause seg faults (#4842)

Also, explicitly listing all the `local` args is error-prone and needs maintenance

### Solution

Just forward all the args from `local_output` to `local`

### DIFF
--- a/local_output/test/test.sh
+++ b/local_output/test/test.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+cd "$(dirname "$0")" || exit
+
+set -euo pipefail
+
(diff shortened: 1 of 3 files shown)


local_output/test/test.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
nicks, line 1: you'll also need to add this to the root test.sh

nicks (approved):
LGTM

PR #223, opened 2021-08-10
extensions: find/run all test.sh scripts rather than listing explicitly

landism:
1. no longer require explicitly adding test.sh scripts to the root test.sh
2. use `make shellcheck` instead of the circleci shellcheck orb:
  a. it was failing due to https://github.com/CircleCI-Public/shellcheck-orb/issues/42
  b. it seems nice to avoid divergence between ci and `make shellcheck` (e.g., `make shellcheck` excludes the ".git-resources" directory and passes "-e SC2001", but AFAICT the shellcheck job does not)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,4 @@
 version: 2.1
-orbs:
-  shellcheck: circleci/shellcheck@2.2.4
 jobs:
   validate:
     docker:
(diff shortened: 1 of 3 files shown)


landism:
well that was more finnicky than I expected...

nicks (approved):

PR #224, opened 2021-08-11
workflow: quality of life tweaks

milas:
Some miscellaneous changes based on (small) hiccups I encountered when using this for the workshop.

I bundled these together just because it was easier and they are pretty small tweaks.

* adjust button labels to be a bit more user friendly / "pretty"
* only reset progress on initial load
  * realized whenever you make a change to `Tiltfile` or any other "config" files, it'd reset you back to beginning
* store temporary files in the `tilt_modules/workflow` directory to avoid creating a bunch of temp files in the repo root
  * undocumented `TILT_WORKFLOW_STATE_PATH` env var can allow setting a custom directory (e.g. tmpdir)
* make sure all temp files meet pattern `workflow_*.tmp` to make it easy to `.gitignore` them
  * see above: hopefully most people already ignore `tilt_modules/` so won't need to, but in the event they do, this makes it easy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 */test/.tilt-cache
 */test/.build
 */test/tilt_modules
+
+workflow_*.tmp
--- a/workflow/README.md
+++ b/workflow/README.md
@@ -7,23 +7,27 @@ workflow(workflow_name : string, resource_name : string, work_cmds=[[]] : [[stri
 ```
 Where `work_cmds` is an array of argv, and `clear_cmd` is a single argv. For example:
 ```
+load('ext://workflow', 'workflow')
+
 workflow(
 	'my-workflow',
-	'my-resource',
-	[ ['echo', '1st'], ['echo', '2nd'], ['echo', '3rd'] ],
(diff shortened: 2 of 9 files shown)


milas:
FYI one file had mixed spaces/tabs and it was making my editor cry, view diff with whitespace changes ignored: https://github.com/tilt-dev/tilt-extensions/pull/224/files?diff=split&w=1

.gitignore
```diff
@@ -2,4 +2,6 @@
 */test/.git-sources
 */test/.tilt-cache
 */test/.build
-*/test/tilt_modules
\ No newline at end of file
+*/test/tilt_modules
+
+workflow_*.tmp
```
nicks, line 7: i also talked to Lucie about this, but I would really like to see these stored in the tilt apiserver as configmaps rather than as random files on disk

nicks (approved):

.gitignore
milas: Yeah - that'd also help it not re-init itself w/o using an env var to hold state. I'll play around with that later and open another PR

PR #228, opened 2021-08-22
kubectl_build: add an easy way to disable kubectl build

glehmann:


kubectl_build/Tiltfile
```diff
@@ -47,3 +53,34 @@ def kubectl_build(ref, context, build_args={}, dockerfile='Dockerfile', live_upd
 
     custom_build(ref, command, [context], disable_push=True, skips_local_docker=True, live_update=live_update,
                  match_in_env_vars=match_in_env_vars, ignore=ignore, entrypoint=entrypoint)
+
+
+def image_build(*args, **kwargs):
```
nicks, line 58: hmmmm...it seems weird to have a function alias?

are you aware that starlark has a way to import functions under a different name?

```
load('ext://kubectl_build', image_build='kubectl_build')
```

kubectl_build/Tiltfile
```diff
@@ -47,3 +53,34 @@ def kubectl_build(ref, context, build_args={}, dockerfile='Dockerfile', live_upd
...
+    """kubectl_build with a less kubectl build obvious name"""
+    kubectl_build(*args, **kwargs)
+
+
+def kubectl_build_enable(v=None):
```
nicks, line 63: can you add docs on this to the README?

nicks (commented):
cool!

kubectl_build/Tiltfile
glehmann: yes I do, but I tried

~~~python
load('ext://kubectl_build', image_build='kubectl_build')
~~~

and

~~~python
load('ext://kubectl_build', 'image_build')
~~~

and liked the latter better in my Tiltfiles.

If you prefer to not have it, I can remove it.

kubectl_build/Tiltfile
glehmann: sure! Forgot to do that :)

kubectl_build/Tiltfile
nicks: :+1: that's fine! I wanted to make sure you knew the feature existed (since not everyone does)

PR #230, opened 2021-08-24
uibutton extension: add text_input

landism:
Update the uibutton extension to allow the user to populate `UIButtonSpec.Inputs`, which was added in https://github.com/tilt-dev/tilt/pull/4857


nicks (approved):

uibutton/README.md
```diff
@@ -100,6 +122,37 @@ cmd_button('svg-btn',
...
+
+When the user clicks the button to run its command, `$NAME` will be set
+to the field's value, e.g.:
+
+```
+Running cmd: echo Hello, $NAME
+Hello, there
+```
+
+Note that if the command needs an env var expanded inside the command itself (e.g., it directly uses `$NAME`, rather than simply invoking a program that uses `$NAME`), it will need to be wrapped in a shell call, e.g. ['sh', '-c', 'mycommand $NAME'].
```
milas, line 154: Ooo yeah this would definitely have tripped me up, good call on documenting 👍

milas (approved):

PR #232, opened 2021-08-24
kubefwd: add a refresh button to restart kubefwd without sudo

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/kubefwd2:

91cbb2b65d65a1f0e82e38b80d5d7fef5be062ba (2021-08-24 15:42:30 -0400)
kubefwd: add a refresh button to restart kubefwd without sudo

### DIFF
--- a/kubefwd/create-refresh-button.sh
+++ b/kubefwd/create-refresh-button.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+set -euo pipefail
+
+TRIGGER="$1"
+
+cat <<EOF | tilt apply -f -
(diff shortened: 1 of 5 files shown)


kubefwd/create-refresh-button.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
landism, line 1: did you consider using the uibutton extension for this? (doesn't matter much for this PR, just wondering if the direction we want to go is every extension that makes a button does something like this or if this was just good enough for now)

landism (approved):

PR #238, opened 2021-08-28
secret: function to generate secret config from a map

johnlayton:
I am currently using Tilt to develop configuration for backstage.
Backstage required Gihtub credentials so I ended up with this;
```
k8s_yaml(secret_from_dict("github", inputs = {
    'AUTH_GITHUB_CLIENT_ID'     : os.getenv("AUTH_GITHUB_CLIENT_ID"),
    'AUTH_GITHUB_CLIENT_SECRET' : os.getenv("AUTH_GITHUB_CLIENT_SECRET")
}))
```
and the proposed function based on the `configmap` extension,


nicks (approved):
LGTM! Nice!!

PR #240, opened 2021-09-01
Helm Remote Extension - Redirect stderr to nul on Windows

TimothyLoyer:
This redirected to stdout on error (such as when no repos were found) which meant that rather than the `|| ver/nul` getting returned, you'd get something like `{"Error": "no repositories to show"}` on stdout which breaks the list comprehension on the next line (note: Helm returns a different error the first time it is run with no repos compared to when you remove all repos- then the command returns an empty list instead).


landism:
LGTM! This just makes the windows behavior match the non-windows behavior, and I don't see any reason they should be different. Thanks!

PR #241, opened 2021-09-02
updated get_local_repo

wejson:
fixed the issue when failing on repos with more than 1 chart

### DIFF
--- a/helm_remote/Tiltfile
+++ b/helm_remote/Tiltfile
@@ -59,7 +59,7 @@ def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], se
     def get_local_repo(repo_name, repo_url):
         # if no repos are present, helm exit code is >0 and stderr output buffered
         added_helm_repos = decode_yaml(local('helm repo list --output yaml 2>/dev/null || true', command_bat='helm repo list --output yaml 2> nul || ver>nul', quiet=True))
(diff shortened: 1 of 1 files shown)


helm_remote/Tiltfile
```diff
@@ -59,7 +59,7 @@ def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], se
     def get_local_repo(repo_name, repo_url):
         # if no repos are present, helm exit code is >0 and stderr output buffered
         added_helm_repos = decode_yaml(local('helm repo list --output yaml 2>/dev/null || true', command_bat='helm repo list --output yaml 2> nul || ver>nul', quiet=True))
-        repo = [item for item in added_helm_repos if item['name'] == chart and item['url'] == repo_url] if added_helm_repos != None else []
+        repo = [item for item in added_helm_repos if item['name'] == repo_name or item['name'] == chart and item['url'] == repo_url] if added_helm_repos != None else []
```
landism, line 62: If I understand correctly, `item['name'] is the name of the repo so we should really just keep the `item['name'] == repo_name` and drop the `item['name'] == chart` bit. What do you think?

I'd guess since `repo_name` defaults to copying `chart`'s value, the difference has probably been immaterial for most users.

wejson: i agree, i kept it cause i wasnt sure how it works for others but come to think about it i dont c any reason for it.
i can remove it if required

landism:
LGTM, thanks!

PR #243, opened 2021-09-02
Maintain LF line-endings for shell scripts on Windows systems

TimothyLoyer:
CRLF line-endings break the ability to execute shell scripts on Windows. This will allow extensions that rely on shell scripts (like Syncback) greater compatibility with Windows/WSL.


landism (approved):
Thanks!

PR #248, opened 2021-09-13
kim: add extension to build images using kim.

lhotrifork:
Hi,

Very simple extension that allows you to build images using kim https://github.com/rancher/kim 
Opens up support for running Tilt against Rancher Desktop https://github.com/rancher-sandbox/rancher-desktop and without having docker installed.

I was having some trouble with my local registry, so I've disabled the `kim push $EXPECTED_REF` command for now, as kim automatically pushes the image into the cluster.

Let me know what you think.
/Lasse

### DIFF
--- a/kim/README.md
+++ b/kim/README.md
@@ -4,15 +4,20 @@ Author: [Lasse Højgaard](https://github.com/lhotrifork)
 
 Use kim (https://github.com/rancher/kim) to build images for Tilt.
 
+## Limitations
+As `kim build` will automatically push images onto kubernetes worker nodes this extension will not run `kim push`.
+
 ## Functions
 
-### `kim_build(ref: str, import_path: str, deps: List[str] = [])`
+### `kim_build(ref: str, import_path: str, deps: List[str] = [], ignore: str, extra_flags?: [str], **kwargs)`
 
 - **ref**: The name of the image to build. Must match the image
    name in the Kubernetes resources you're deploying.
-- **import_path**: A Go import path of the binary to build.
-- **deps**: A list of dependencies that can trigger rebuilds.
-
+- **context**: The build context of the image to build. Expressed as a file path.
+- **ignore**: Changes to the given files or directories do not trigger rebuilds.
+      Does not affect the build context.
+- **extra_flags**: Extra flags to pass to kim build.
+- **kwargs: will be passed to the underlying `custom_build` call
 ## Example Usage
 
 ```
--- a/kim/README.md
+++ b/kim/README.md
@@ -2,26 +2,23 @@
 
 Author: [Lasse Højgaard](https://github.com/lhotrifork)
 
-Use kim (https://github.com/rancher/kim) to build images for Tilt.
+Use [kim](https://github.com/rancher/kim) to build images for Tilt.
 
 ## Limitations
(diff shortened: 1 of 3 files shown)


lhotrifork:
Hi @victorwuky,
Do you have time for a review? :)

/Lasse

kim/README.md
```diff
@@ -0,0 +1,22 @@
+# kim
+
+Author: [Lasse Højgaard](https://github.com/lhotrifork)
+
+Use kim (https://github.com/rancher/kim) to build images for Tilt.
+
+## Functions
+
+### `kim_build(ref: str, import_path: str, deps: List[str] = [])`
```
milas, line 9: Can you update the parameters and their docs to match the function signature?

kim/Tiltfile
```diff
@@ -0,0 +1,29 @@
...
+  custom_build(
+    ref=ref,
+    command=(
+      "set -ex\n" +
+      "kim build -t $EXPECTED_REF %s %s\n" +
+      push_cmd
+    ) % (extra_flags_str, shlex.quote(context)),
+    ignore=ignore,
+    deps=[context],
+    skips_local_docker=True,
```
milas, line 28: You can set `disable_push=True` here and omit the "already pushed $EXPECTED_REF" message.

Right now, since `kim` is experimental, I think its usage is limited to Rancher Desktop (or other single node K8s dev clusters), where push is unnecessary, so this seems fine. However, it would be good to add a note to the `README` noting this caveat.

kim/Tiltfile
```diff
@@ -0,0 +1,29 @@
...
+  extra_flags = extra_flags or []
+  extra_flags_str = ' '.join([shlex.quote(f) for f in extra_flags])
+
+  push_cmd = "echo already pushed $EXPECTED_REF"
+
+  custom_build(
+    ref=ref,
+    command=(
+      "set -ex\n" +
+      "kim build -t $EXPECTED_REF %s %s\n" +
```
milas, line 23: I think the "already pushed" message is redundant (see below), but can you also remove the `set -ex`? Tilt will fail the build as-is if `kim` returns a non-zero exit code and it shows the command it's going to run, so this is redundant.

But more importantly, making this _just_ `kim build ...` will make sure it works on Windows as well without a separate `command_bat`.

kim/Tiltfile
```diff
@@ -0,0 +1,29 @@
+
+# -*- mode: Python -*-
+
+def kim_build(ref, context, ignore=None, extra_flags=None):
```
milas, line 4: It'd be nice to add `**kwargs` to pass through to `custom_build` so that you can specify `live_update` etc.

milas (commented):
Thanks for the PR! I left a few minor comments to polish things up, but no major concerns 😄

lhotrifork:
Hi @milas,
Thanks for the review! I'm not that familiar with python, so let me know if you have any more edits.
I'll squash the commits when you're satisfied.

I saw that somebody wrote a blogpost implementing exactly this extension :D 
https://blog.tilt.dev/2021/09/07/rancher-desktop.html

milas:
Thanks for the changes! I just pushed a couple tweaks - there was an errant `+` on the command line causing a syntax error and the args in the README were slightly out of sync. Squashing + merging now.

PR #252, opened 2021-09-18
namespace Add labels and annotation options to namespace_create and namespace_yaml

duck1123:
This commit adds optional parameters to the namespace functions to set labels and annotations at namespace creation time. If either parameter is not set, the corresponding section is excluded from the yaml.


nicks (approved):
LGTM! Nice!

Issue #259, opened 2021-10-04
Allow deps to be different from context when using podman_build extension


ElArtista:
Currently, the podman_build extension binds the `deps` param in the custom_build function to passed the context:
https://github.com/tilt-dev/tilt-extensions/blob/53b333d1ea91dc9452229b689f8242c34189162d/podman/Tiltfile#L30

This is not always the case, as most of the times we want to observe just the files needed by the image to trigger a rebuild.
We should add an extra deps param that defaults to the context if not passed.

PR #260, opened 2021-10-04
Allow deps to be different to context in podman_build extension

ElArtista:
Relevant to #259


landism (changes requested):
Can you move `deps` to be the last arg? Otherwise, this changes the meaning of calls w/ positional args. E.g., in `podman_build('foo', '.', ['.git'])`, `.git` would be a `dep` rather than an `ignore`

ElArtista:
I was trying to be as closely as possible to the custom_build command which specifies the `deps` param before the `ignore` param. Are you sure this is what we do want?

PR #278, opened 2021-10-27
syncback: krsync.sh uses existing rsync

nicksieger:
In our apps we are running container processes as non-root so the attempt to upload a custom rsync binary fails due to not being able to write to `/bin`. The workaround I used locally was to include rsync in our base images and remove that code from krsync.sh.

So that I can stop using the workaround, this PR attempts to detect if a compatible rsync is already installed and will use it instead.


nicks:
This code has historically had a lot of bugs. I'm not confident that we can safely make a change of this magnitude/complexity without adding some tests.

syncback/krsync.sh
```diff
@@ -5,41 +5,60 @@
...
+install_rsync_tilt() {
+    if [ -z "$RSYNC_PATH" ]; then
+        # rsync already checked/installed
+        return 0
+    fi
+
+    local tar_path_local=${0%/*}/rsync.tilt.tar  # TODO: option for windows binary
+    # ensure our rsync bin is present on container in the expected location
+    set +e # disable exit on error
+    kubectl exec $K8S_OBJECT -- ls "$rsync_tilt_path" > /dev/null 2>&1
```
nicks, line 22: do you need to quote "$K8S_OBJECT" like it was before?

syncback/krsync.sh
```diff
@@ -5,41 +5,60 @@
...
+    exit_code="$?"
+    set -e
+    if [[ "$exit_code" != "0" ]]; then
+        # shellcheck disable=SC2002
+        cat "$tar_path_local" | kubectl exec -i $K8S_OBJECT -- tar -xf - -C $(dirname $rsync_tilt_path)
```
nicks, line 27: ditto

nicks (approved):
LGTM!

syncback/krsync.sh
nicksieger: Nope! That's part of the fix to #209. It comes into the script as a single argument (quoted from the Tiltfile) but we don't quote it so that kubectl interprets the namespace and container args separately.

nicksieger: I'll make a note in the script.

PR #281, opened 2021-11-02
nix: added build_nix_image

afreakk:
Added function to build images with nix.
I used `sed` to extract image-name from `docker load`, it kindof relies on docker not changing output from `docker load`.
Without that part, user would have to pass in the full name of the docker image  being created by nix.
Which is optional in nix, you dont have to supply a name, which will cause it to have autogenerated name.
So its nice if we can capture it, and not sure if there is a better way.


nicks (approved):
LGTM! This is so cool!

Issue #287, opened 2021-11-08
Add reference for extension: docker_build_with_restart()


Chimney42:
During exterminator duties, I tried to find the docs on the extension `docker_build_with_restart`.
It was quite buried, and it took a while for me to find it through this blogpost: https://docs.tilt.dev/live_update_reference.html#restarting-your-process

If there is a function you can call in the Tiltfile, there should be an easy way to find the docs by a simple search, e.g. in the readme of this repository, it could link to the `restart_process` docs.

PR #288, opened 2021-11-08
Enhance docs docker restart

Chimney42:
Example solution for: https://github.com/tilt-dev/tilt-extensions/issues/287

I think it should be easy to find the docs on any function I can call in a Tiltfile.


nicks (approved):
fine with me, tho i feel like we're getting to a point where we need to be auto-generating these

PR #292, opened 2021-11-10
nix: fix sed expression

afreakk:
Noticed some issues with the previous extraction of `$IMG_NAME` from `docker load`.
The stdout filtering/sed only worked when the image had been previously loaded, because then the stdout only contained a single line, otherwise docker printed multiple lines to stdout, so `$IMG_NAME` var got multiple lines, which resulted in the tagging process not working.
Now I use `-n` which suppress automatic printing of pattern space.
Then append `p` which prints the current pattern space, which results in only printint output from matching line.

### DIFF
--- a/nix/Tiltfile
+++ b/nix/Tiltfile
@@ -6,9 +6,9 @@ def build_nix_image(ref, path = "", attr_path = "", deps = []):
 
     #hopefully docker does not change output from docker load
     #this sed expression extracts the name from docker load stdout
-    sed_expr = r"s@Loaded image: \(\S*\)@\1@"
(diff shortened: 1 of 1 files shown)


landism (approved):
This change looks good to me. Thanks!

FWIW, if you want to avoid relying on `docker load` output, you could consider something like this:
```
commands = [
  'IMG_PATH="$({})"'.format(build_cmd),
  'docker load --input "$IMG_PATH"',
  'IMG_NAME=$(tar -Oxf "IMG_PATH" manifest.json | jq -r ".[0].RepoTags[0]")',
  'docker tag $IMG_NAME $EXPECTED_REF',
]
```

PR #296, opened 2021-11-13
nix: get tag from manifest.json

afreakk:
- get tag from `manifest.json` inside `tar.gz` instead of `docker load` stdout.
- `set -e` to have commands fail early to get easier to read errors.
- `--no-out-link` so the built `.tar.gz` will be collected by nix garbage collector.

After tips from https://github.com/tilt-dev/tilt-extensions/pull/292#pullrequestreview-803039132


nicks (approved):
thanks! LGTM

nicks:
(sorry for the delay, didn't see this initially)

PR #304, opened 2021-12-08
Fix dockerfile argument of docker_build_sub()

remram44:
The 'dockerfile' argument needs to be passed to the first `docker_build()` but not the second, otherwise this happens:
```
Error: Cannot specify both dockerfile and dockerfile_contents keyword arguments
```


nicks:
i don't think this bug fix is quite right, but i understand the underlying bug. gmas!

nicks:
https://github.com/tilt-dev/tilt-extensions/pull/306

nicks:
closing this as obselete. thanks!

PR #305, opened 2021-12-10
k8s_attach: an extension that can monitor and live_update existing servers

nicks:
Hello @milas,

Please review the following commits I made in branch nicks/attach:

e055d45cd5c84b060e7286dbfb871ed9dd9c5ea1 (2021-12-10 09:29:28 -0500)
k8s_attach: an extension that can monitor and live_update existing servers

### DIFF
--- a/k8s_attach/Tiltfile
+++ b/k8s_attach/Tiltfile
@@ -1,0 +1,28 @@
+
+def k8s_attach(name, obj, namespace="", deps=[], live_update=None, **kwargs):
+  """Attach to a kubernetes resource.
+
+  Arguments:
+  name: The name for the resource in the UI.
+  obj: An object name in the form "kind/name", as you would express it to kubectl. e.g., deployment/my-deployment.
+  namespace: The namespace of the object. If not specified, uses the default namespace.
+  live_update: Live-update rules to copy files and run commands in the server whenever
+    they change locally.
+  deps: When using live-update, you must also specify the files as a dependency.
+  **kwargs: Arguments to pass to the underlying k8s_resource, including labels.
+  """
+
+  args = ["kubectl", "get", "-o=yaml", obj]
+  if namespace:
+    args.append("-n", namespace)
+
+  k8s_custom_deploy(
+    name,
+    apply_cmd=args,
+    delete_cmd=["echo", "ok"],
+    deps=deps,
+    live_update=live_update,
+    image_selector='busybox')
+
+  k8s_resource(
+    name, **kwargs)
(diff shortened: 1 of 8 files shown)


k8s_attach/test/Tiltfile
```diff
@@ -0,0 +1,15 @@
...
+# outside of the main
+local('kubectl apply -f deployment.yaml')
+
+k8s_attach(
+  'example-html',
+  'deployment/example-html',
+  live_update=[
+    sync('./', '/src'),
+  ],
+  deps=['./'],
```
nicks, line 14: @milas i wonder if we could auto-populate deps, rather than force it to be specified twice?

k8s_attach/Tiltfile
```diff
@@ -0,0 +1,28 @@
...
+  if namespace:
+    args.append("-n", namespace)
+
+  k8s_custom_deploy(
+    name,
+    apply_cmd=args,
+    delete_cmd=["echo", "ok"],
+    deps=deps,
+    live_update=live_update,
+    image_selector='busybox')
```
milas, line 25: I think this needs to be passed through from an arg to `k8s_attach`, no? (And we might also want to support `container_selector`)

k8s_attach/test/Tiltfile
milas: Yeah, on the extension side, it'd be nice if the LU Starlark objects were structs so you could read `.source_path` off of them to build up a deps list from the rules (e.g. this would also be handy for `file_sync_only`). Though this probably isn't worth the effort (might be better served by e.g. figuring out how to more deeply integrate the API codegen'd variants)

For `k8s_custom_deploy` specifically, we could also do this on the Go side during assembly - I thought about this originally but wasn't sure if it was worth different semantics than `docker_build` (even though they're _very_ different, their APIs are quite similar). And then `deps` could become `apply_deps=[]` (optional) which makes more sense. WDYT?

k8s_attach/Tiltfile
```diff
@@ -0,0 +1,28 @@
...
+  """
+
+  args = ["kubectl", "get", "-o=yaml", obj]
+  if namespace:
+    args.append("-n", namespace)
+
+  k8s_custom_deploy(
+    name,
+    apply_cmd=args,
+    delete_cmd=["echo", "ok"],
```
milas, line 22: Maybe "Skipping delete of $obj, which is managed outside Tilt" or something along those lines?

milas (changes requested):
Nice! @hyu and I were actually talking about functionality like this last week as well! 🎉 

Only issue I see is the hardcoded `busybox` image selector.

k8s_attach/Tiltfile
nicks: done

k8s_attach/Tiltfile
nicks: done

k8s_attach/test/Tiltfile
nicks: ya tbh we should make sync an alias of LiveUpdateSync and use the autogenerated objs

milas (approved):

PR #312, opened 2021-12-16
kubefwd/ngrok: use portable mktemp invocation for macOS/Linux

milas:
BSD `mktemp` has some differences compared to the GNU variant,
so this is the most portable way to make temporary files that
works across macOS + Linux that I could find. Note that the ngrok
file won't end in `.yml` anymore because BSD `mktemp` only allows
the randomized part of the template (i.e. `XXXXXX`) to be at the
END of the filename.

Both of these assume a POSIX-y env as-is (e.g. they're running
Bash scripts for their main operator functionality, so no attempt
at supporting Windows cmd is made here.

### DIFF
--- a/kubefwd/Tiltfile
+++ b/kubefwd/Tiltfile
@@ -5,7 +5,7 @@ cfg = config.parse()
 
 namespaces = cfg.get('namespaces', [])
 
-trigger_file = str(local('mktemp --suffix -tilt-kubefwd')).strip()
+trigger_file = str(local('mktemp "${TMPDIR:-/tmp}/tilt-kubefwd.XXXXXXXXX"')).strip()
 
 # Creates a button that refreshes the kubefwd without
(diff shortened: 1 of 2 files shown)


nicks (approved):

nicksieger:
LGTM also. I've been wondering if we should make a builtin `temp_file(pattern)` and `temp_dir(pattern)` functions that put the new items in [the `scratchDir` field on the tiltfile state](https://github.com/tilt-dev/tilt/blob/master/internal/tiltfile/tiltfile_state.go#L146-L148) and start using those instead.

nicksieger (approved):

milas:
💯 for tmp helpers, trying to get portable invocations across macOS/Linux is awful (to say nothing of Windows 😭) and having Tilt clean them up would be cherry on top

PR #317, opened 2022-01-05
helm_remote: add a new extension for installing with helm natively

nicks:
Hello @milas, @nicksieger,

Please review the following commits I made in branch nicks/resource:

f08eeb7558fdfcfa7aad5a140bd3b070b9a1523e (2022-01-05 17:34:18 -0500)
helm_remote: add a new extension for installing with helm natively


nicks:
(don't worry, this is much smaller than it looks, this is mostly generated by `helm create`)

this fixes https://github.com/tilt-dev/tilt/issues/4375 and should also address a number of other helm bugs in the main tilt repo.

nicksieger:
LGTM, a thought though - could the test run `helm create` so we can track helm changes a bit better and the example doesn't bit rot? I realize it's a trade-off to having a fully working example committed to source control.

nicksieger (approved):

nicks:
ya, i understand where you're coming from -- but people tend to look at these as working examples. i think they would be super confused if they saw us running `helm create` dynamically, since it's so normal to run `helm create` and checkin the result. 

Even if the output of `helm create` changed, you would still have tons and tons of old charts, and you'd still need to know how to configure tilt to run them.

PR #319, opened 2022-01-11
honeycomb: support custom_build

jazzdan:
Ends up looking something like this:

Screen Shot 2022-01-10 at 9 47 54 PM


jazzdan:
Looks like the tests aren't running against the latest Tilt?

nicks:
ya - we have to bump the version manually. tho i think you'll still need a try...catch block so that this doesn't break older users (i think CmdImage was added VERY recently, like the last release or two)

nicks:
added some tests and sent out https://github.com/tilt-dev/tilt-extensions/pull/329

jazzdan:
Whoops, thanks for running with this @nicks! It got lost in the shuffle of moving.

nicks:
superseded by https://github.com/tilt-dev/tilt-extensions/pull/329 !

PR #321, opened 2022-01-12
uibutton: fix errors running some commands

milas:
Only create buttons for `tilt up` and `tilt ci`. Other commands,
such as `tilt alpha tiltfile-result`, load the Tiltfile but do NOT
launch the apiserver, so buttons can't be created.


lianmakesthings (approved):
Tested and works 👍

PR #322, opened 2022-01-14
color: add extension for ANSI codes in Tiltfile

milas:
Add ANSI color codes to your Tiltfile log messages.

Works with Tiltfile built-ins such as `print`, `warn`, `fail`, `exit`
in addition to custom functions.

screenshot of colorized output from Tiltfile logging in CLI

screenshot of colorized output from Tiltfile logging in web UI

### DIFF
--- a/color/Tiltfile
+++ b/color/Tiltfile
@@ -5,7 +5,7 @@ RED='\033[0;31m'
 GREEN='\033[0;32m'
 YELLOW='\033[0;33m'
 BLUE='\033[0;34m'
-magenta='\033[0;35m'
+MAGENTA='\033[0;35m'
 CYAN='\033[0;36m'
 WHITE='\033[0;37m'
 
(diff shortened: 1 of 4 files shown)


color/Tiltfile
```diff
@@ -0,0 +1,74 @@
+RESET='\033[0m'
+
+BLACK='\033[0;30m'
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[0;33m'
+BLUE='\033[0;34m'
+magenta='\033[0;35m'
```
nicksieger, line 8: How come this one isn't capitalized?

nicksieger (approved):

lianmakesthings (approved):

color/Tiltfile
milas: Because I don't know how to use find and replace 🤦 

(Good catch 😅)

Issue #324, opened 2022-01-20
helm_resource: add support for namespaces


markdingram:
Currently namespace can be provided as one of the flags. However this then isn't passed along to the other helm invocations that follow the install.

(I'll submit a PR?)

milas:
Thanks for the report!

Only global Helm flags will be acceptable for both `helm install` and `helm get manifest`, so I think we'll either need a `global_flags` or an explicit `namespace` arg.

We happily take PRs! (I'll try to get to this tomorrow otherwise 🙃)

markdingram:
fixed by #327

PR #327, opened 2022-01-21
helm_resource: add support for helm namespace (#324)

markdingram:


helm_resource/README.md
```diff
@@ -61,6 +62,8 @@ chart. Must be the same length as `image_deps`.  There are two common patterns.
 
 `flags`: Additional flags to pass to `helm install` (e.g., `['--set', 'key=value']`)
 
+`global_flags`: Additional flags to pass to all helm invocations (e.g., `['--namespace', 'app']`)
```
nicks, line 65: are there any flags that you need beside `--namespace`? `app` isn't a flag...

feels like once you get into the business of parsing the flags (which is going to be tricky and error-prone), we might as well have a dedicated `namespace` argument

markdingram: this was to match up with the global flags section as shown`helm install --help` (and other helm commands)
```
...
Global Flags:
      --debug                       enable verbose output
      --kube-apiserver string       the address and the port for the Kubernetes API server
      --kube-as-group stringArray   group to impersonate for the operation, this flag can be repeated to specify multiple groups.
      --kube-as-user string         username to impersonate for the operation
      --kube-ca-file string         the certificate authority file for the Kubernetes API server connection
      --kube-context string         name of the kubeconfig context to use
      --kube-token string           bearer token used for authentication
      --kubeconfig string           path to the kubeconfig file
  -n, --namespace string            namespace scope for this request
      --registry-config string      path to the registry config file (default "/Users/markingram/Library/Preferences/helm/registry.json")
      --repository-cache string     path to the file containing cached repository indexes (default "/Users/markingram/Library/Caches/helm/repository")
      --repository-config string    path to the file containing repository names and URLs (default "/Users/markingram/Library/Preferences/helm/repositories.yaml")
```


my usecase is for namespace though, so happy to switch

helm_resource/README.md
```diff
@@ -47,6 +48,9 @@ reference.
 
 `release_name`: The name of the release. If not specified, defaults to the Tilt UI resource name.
 
+`namespace`: Install the release into the specified namespace. Also, when specified, the chart
+may only create objects in that particular namespace.
+
```
markdingram, line 53: Calling out an edge case seen when trying to set a namespace on a chart that produces objects in multiple namespaces. 

This doesn't work as the kubectl command will lead to errors like `the namespace from the provided object "abc" does not match the namespace "xyz". You must pass '--namespace abc' to perform this operation.`.

helm_resource/README.md
markdingram: PR uses dedicated `namespace` argument now.

nicks (approved):
nice! LGTM!

PR #330, opened 2022-01-21
helm_resource: fix sed errors on old macOS

### LINKED ISSUE #328: helm_resource: sed error on OSX Catalina
The use of sed [here](https://github.com/tilt-dev/tilt-extensions/blob/fb6856b332bcb4f257a2f50e4a0c7f255d85dc55/helm_resource/Tiltfile#L76) erros on OSX Catalina (10.15.7) 

```
❯ sed -r
sed: illegal option -- r
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
```

The version of sed on Monterey (12.1) does now include -r.

Problem is seen on a work machine, so upgrading isn't an option right now - there is a workaround by tweaking the path to use <https://formulae.brew.sh/formula/gnu-sed>.

milas:
Old versions of BSD `sed` don't include the `-r` flag for extended
regex (they instead use `-E`). This prevents things from working
on older macOS versions (e.g. 10.15).

Use `-e` instead for greater portability, which works fine for our
regexes with a slight extra escape on the capture group for BRE
compatibility.

Fixes #328.

### DIFF
--- a/helm_resource/Tiltfile
+++ b/helm_resource/Tiltfile
@@ -73,8 +73,8 @@ def helm_resource(
       # NOTE(nick): This is a PITA because it's hard to separate
       # out the repository from the tag.
       flags.extend([
-        '--set', '%s=$(echo "$TILT_IMAGE_%s" | sed -r \'s,:[^/]*$,,\')' % (shlex.quote(key[0]), i),
-        '--set', '%s=$(echo "$TILT_IMAGE_%s" | sed -r \'s,^.*:([^/]*)$,\\1,\')' % (shlex.quote(key[1]), i),
+        '--set', '%s=$(echo "$TILT_IMAGE_%s" | sed -e \'s,:[^/]*$,,\')' % (shlex.quote(key[0]), i),
+        '--set', '%s=$(echo "$TILT_IMAGE_%s" | sed -e \'s,^.*:\\([^/]*\\)$,\\1,\')' % (shlex.quote(key[1]), i),
       ])
     else:
       fail("invalid argument to image_keys at %s: %s" % (i, type(key)))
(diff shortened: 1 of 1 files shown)


milas:
I tested this across macOS (12.1) / Alpine / busybox / Ubuntu to be safe.

helm_resource/Tiltfile
```diff
@@ -73,8 +73,8 @@ def helm_resource(
       # NOTE(nick): This is a PITA because it's hard to separate
       # out the repository from the tag.
       flags.extend([
-        '--set', '%s=$(echo "$TILT_IMAGE_%s" | sed -r \'s,:[^/]*$,,\')' % (shlex.quote(key[0]), i),
-        '--set', '%s=$(echo "$TILT_IMAGE_%s" | sed -r \'s,^.*:([^/]*)$,\\1,\')' % (shlex.quote(key[1]), i),
+        '--set', '%s=$(echo "$TILT_IMAGE_%s" | sed -e \'s,:[^/]*$,,\')' % (shlex.quote(key[0]), i),
+        '--set', '%s=$(echo "$TILT_IMAGE_%s" | sed -e \'s,^.*:\\([^/]*\\)$,\\1,\')' % (shlex.quote(key[1]), i),
```
nicks, line 77: do you need the `-e` at all? i think this _might_ work with the "default" regexp mode (though i could be wrong)

nicks (approved):

helm_resource/Tiltfile
milas: Yeah, it's unnecessary - dropped in 1c01180

PR #334, opened 2022-02-09
honeycomb: fix bug where only initial builds would be reported

jazzdan:
Prior to this change we were always sending the current time to the honeycomb_events script instead of the last_report_time. This meant that the "last_report_time" from the honeycomb_event script's perspective was always after whatever builds may have taken place, and so it never reported any builds except for the first time it was executed when it wasn't passed a "last_report_time".


nicks (approved):
LGTM!

jazzdan:
Thanks @nicks!

PR #341, opened 2022-02-16
knative: increase timeout to fix delete (and test)

mjgallag:


mjgallag:
@nicks my circeci tests are running now, seems a bit weird i'd need a circleci user setup and connected to github myself, but i'll leave it as is for now in case that fixed it!

nicks (approved):
oof, i wonder why knative takes so long to tear down? in any case, LGTM!

mjgallag:
@nicks I haven't been able to figure it out, not that I've dug super deep, but I've been working with knative intermittently for a few years and it's unfortunately always had this issue.

PR #347, opened 2022-02-18
docker_run: new extension/function to run standalone containers

nicksieger:
With simultaneous docker compose+k8s orchestration now available, I think the time is right for this to exist, to allow people to more idiomatically manage simple docker container builds+runs within the Tiltfile.

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -29,3 +29,9 @@ RUN curl -sSL "https://github.com/buildpacks/pack/releases/download/v0.21.1/pack
 RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
   echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
   apt update && apt install -y yarn
+
+# install docker-compose
+ARG DOCKER_COMPOSE_VERSION=v2.2.3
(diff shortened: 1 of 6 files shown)


nicks (approved):
one thing that's worth calling out:

tilt merges all docker-compose files into a single docker-compose application.

docker-compose has some surprising behavior in how it handles merges - e.g., using the working directory of the first file, merging data structured together, etc. https://github.com/compose-spec/compose-spec/blob/master/spec.md#compose-file. It doesn't work like Kubernetes, where different objects are independent.

milas (commented):
Interesting approach using the built-in Docker Compose support! (It makes me wonder if we should really allow `local_resource` to have optional `image_deps` and inject the built image names via env vars, similar to how we do for `custom_build` / `k8s_custom_deploy` now.)

At a high-level, maybe it's just me, but I see the big value of an extension/built-in like this to be for running Dockerized "tasks" (e.g. closer to Makefile targets), so the emphasis on more daemon tasks is a little odd - it's not clear to my why I'd do this instead of writing a Docker Compose YAML file and importing that with Tilt.

Also, fwiw for "tasks", I'd probably want to be able to set up a bind mount (or perhaps ideally, have Tilt transparently create a volume and perform two-way sync using it + its filewatch logic so that it works even with remote Docker hosts).

I'm also a tiny bit afraid folks are going to think there's some serious wizardry/magic going on here and that these "server" containers are going to be able to interop with e.g. K8s resources in terms of ports etc.

milas (approved):
(oops sorry previous review was meant to be an approval - despite how it might have seemed from all my rambling thoughts 😅)

PR #350, opened 2022-02-23
syncback: make sure the krsync entrypoint does not need escaping

nicks:
Hello @nicksieger,

Please review the following commits I made in branch nicks/syncback:

bd46d60bf70723e4c08a4390078bf45a4d733214 (2022-02-23 10:34:02 -0500)
syncback: make sure the krsync entrypoint does not need escaping


nicks:
i think there might be an underlying rsync bug here. Anything passed to `rsync --rsh` did not properly escape its arguments. but we're already using it in kind of a bizarre way.

nicksieger (approved):
The fun continues! 🤡

PR #353, opened 2022-03-01
syncback: Fix paths not being properly applied

NanoSector:
The following snippet from the README would sync __everything__ from the container at the specified path:

```python
    syncback('syncback-data', 'job/data-cron',
             '/data/',
             ignore=['k8s.yaml', 'config.json']
   )
```

According to answers on [stack overflow](https://stackoverflow.com/questions/11111562/rsync-copy-over-only-certain-types-of-files-using-include-option) the include option overrides the exclude option, so it's mandatory to first exclude everything before including things.

PR #355, opened 2022-03-03
syncback: fix paths/include/exclude behavior

nicksieger:
- Fix paths not being properly applied
- syncback: use `paths` in test
- syncback: fix use of paths/include/exclude


NanoSector (approved):
LGTM, your tests show that the example.txt's don't get synced back so I trust that my issue is resolved.

I can test in the office tomorrow if you want extra verification. I'll also close #353 considering you've inherited those changes in this PR.

NanoSector:
I just tested this at the office with my sample project and it seems to work fine now :+1:

PR #356, opened 2022-03-04
helm_resource: Add port_forwards parameter for underlying k8s_resource

joshjennings98:
We have an API server running using tilt that we wish to communicate with. The server listens on a specific port but the underlying `k8s_resource()` does not let us specify a port meaning that when we send requests we don't know which port to send them to. 

By adding the `port_forwards` argument we can optionally set the port on the `k8s_resource` function so we know to which port we can send requests to the server.

### DIFF
--- a/helm_resource/Tiltfile
+++ b/helm_resource/Tiltfile
@@ -107,10 +107,14 @@ helm uninstall %(namespace_arg)s %(release_name)s || true
     live_update=live_update)
 
   if resource_deps:
-    k8s_resource(name, resource_deps=resource_deps, port_forwards=port_forwards)
+    k8s_resource(name, resource_deps=resource_deps)
 
   if labels:
-    k8s_resource(name, labels=labels, port_forwards=port_forwards)
+    k8s_resource(name, labels=labels)
+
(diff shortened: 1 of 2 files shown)


helm_resource/Tiltfile
```diff
@@ -105,10 +107,10 @@ helm uninstall %(namespace_arg)s %(release_name)s || true
     live_update=live_update)
 
   if resource_deps:
-    k8s_resource(name, resource_deps=resource_deps)
+    k8s_resource(name, resource_deps=resource_deps, port_forwards=port_forwards)
```
nicks, line 110: can you change this to:

```
if len(port_forwards):
  k8s_resource(name, port_forwards=port_forwards)
```

otherwise, they might get added twice, which is not what you want

nicks (commented):
thanks for this! one minor bugfix suggestion

helm_resource/Tiltfile
joshjennings98: Thank you for pointing that out

nicks (approved):

PR #359, opened 2022-03-08
syncback: split rsync callback off from the main script

nicks:
Hello @nicksieger,

Please review the following commits I made in branch nicks/syncback:

6f8dcacde94c5196b694bba5865d8aecd53e3c39 (2022-03-08 16:06:25 -0500)
syncback: split rsync callback off from the main script


nicks:
unless i'm missing something, there's not really a good reason for this script to be self-referential?

nicksieger (approved):
Agreed. The script from stack overflow was being a little too clever by half, this makes the flow clearer.

nicksieger:
The tests always succeed for me locally. Maybe we add a `set +x; ... set -x` around the rsync version detection commands; that seems to be what's occasionally failing.

nicks:
ya, i added some `set -x` calls to help diagnose on CI

nicks:
OH! I know the bug. The problem is there's a race condition - we don't know when the `apk add` will finish. so i added a readiness probe for that

PR #361, opened 2022-03-09
syncback: another attempt at fixing CI errors

nicks:
Hello @nicksieger,

Please review the following commits I made in branch nicks/syncback2:

007a70a71b889e8d0f57caa28b5b4cbd5bb478cb (2022-03-09 12:44:07 -0500)
syncback: another attempt at fixing CI errors

### DIFF
--- a/syncback/krsync.sh
+++ b/syncback/krsync.sh
@@ -5,6 +5,12 @@
 
 set -e
 
+if [ "$CI" != "" ]; then
+    # Print syncback output on CI to make it easier
+    # to diagnose failures.
+    set -x
+fi
+
 # make sure local tmp dir exists (see -T= argument in Tiltfile)
 mkdir -p /tmp/rsync.tilt
 mkdir -p /tmp/rsync.tilt.krsync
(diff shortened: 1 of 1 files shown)


nicks:
OK! Here's my theory of what's going on! (You may know more about bash pipes internals than me, so please affirm / deny this if it sounds plausible and/or nuts)

- kubectl creates a pipe between the remote session and the local terminal
- `ls` starts writing to the pipe
- `head -1` reads the first line

BUT, depending on the pipe internals, `ls` may or may not hang, because it's waiting on the pipe to clear up. Eventually `kubectl` kills it with a SIGPIPE (141), which gets treated as an error (even though it succeeded)

syncback/krsync.sh
```diff
@@ -29,19 +35,20 @@ install_rsync_tilt() {
...
 find_rsync_path() {
     local rsync_path=--rsync-path=$rsync_tilt_path
-    # ensure pipeline status is non-zero if rsync is missing in the container
-    set -o pipefail
     # see if we already have a compatible rsync on the container
-    detected_rsync_version=$(kubectl exec $K8S_OBJECT -- rsync --version 2> /dev/null \
-                              | head -1 | awk '{ print $3 }')
-    if [ $? -eq 0 -a "$detected_rsync_version" ] && [[ "$detected_rsync_version" > 3 ]]; then
-        rsync_path=
+    rsync_output=$(kubectl exec $K8S_OBJECT -- rsync --version 2> /dev/null)
```
nicksieger, line 46: In [this previous run](https://app.circleci.com/pipelines/github/tilt-dev/tilt-extensions/943/workflows/037f4dfe-a1cf-4a94-afa8-372189868715/jobs/2088?invite=true#step-103-2135) it looks like kubectl may have returned a non-zero exit code even though it looks like rsync wrote output to the pipe. I wonder if we should disregard kubectl exit status entirely and go back to do something like 
```
detected_rsync_version=$(kubectl -v=0 exec $K8S_OBJECT -- rsync --version 2> /dev/null | grep rsync | head -1 | awk '{ print $3 }')
if [ "$detected_rsync_version" ] && [[ "$detected_rsync_version" > 3 ]]; then
```

nicksieger: or 
```
detected_rsync_version=$(kubectl -v=0 exec $K8S_OBJECT -- rsync --version 2> /dev/null | grep "rsync *version" | awk '{ print $3 }')
```

nicks: ya, the previous run is from before the previous PR, that didn't have the pipe changes.

i'm not totally convinced the grep solutions will work right, since they have the same pipe problems.

nicksieger (approved):
Ok, let's go with this then! Thanks for the sleuthing.

PR #366, opened 2022-03-21
ci: upgrade to Go 1.18

milas:


milas:
> looks like you need to upgrade the validate script

I knew I would miss a `go get` -> `go install` somewhere during this 😬

nicksieger (commented):
Don't you need to bump the version in go.mod?

milas:
> Don't you need to bump the version in go.mod?

Guessing you mean on the Tilt proper side? FWIW my thinking of not bumping it from `go 1.17` over there was to generally try to provide a bit of a transition period where we _require_ `v(N-1)` but target (i.e. compile with) `vN`. The 1.18 transition was thankfully pretty painless, but for example, Homebrew has struggled in the past to get the new Go release out quickly [due to package compatibility issues], so I think it's a generally good idea from a process standpoint to follow. That said, Homebrew has 1.18 out now, and it _was_ a low friction release, so I feel fine bumping `go.mod` version sooner (i.e. now 😛)

nicksieger:
Oh yeah, I did. Thanks.

PR #373, opened 2022-04-02
Fix issue with multi-part target in child build parameters

engson:
With parent as a multi-part dockerfile, it is often useful to build with a dockerfile target. This target should not be passed down to child dockerfile build, as it will always fail with the selected target stage not being available to child.


nicks (approved):
LGTM! thanks!

Issue #368, opened 2022-03-30
kubefwd: executing script with spaces in path


erickhun:
It seems like on mac, there is issue with interpreting space in a path when trying to execute `run-kubefwd.sh`

The `kube:fwd` ouptut:
```
Running cmd: ./sudo-kubefwd.sh /var/folders/xh/003_hq_j3qq29mtv3nn8kw6c0000gn/T//tilt-kubefwd.nXOFvvLqd
+ /usr/bin/osascript -e 'do shell script "/Users/eric/Library/Application Support/tilt-dev/tilt_modules/github.com/tilt-dev/tilt-extensions/kubefwd/run-kubefwd.sh /opt/homebrew/bin/kubefwd /Users/eric/.kube/config /opt/homebrew/bin/entr /var/folders/xh/003_hq_j3qq29mtv3nn8kw6c0000gn/T//tilt-kubefwd.nXOFvvLqd" with administrator privileges'
0:315: execution error: /bin/sh: /Users/eric/Library/Application: is a directory (126)
```

I've tried to add a backlash in the path when there is a space in the [`sudo-kubefed.sh`](https://github.com/tilt-dev/tilt-extensions/blob/master/kubefwd/sudo-kubefwd.sh#L12-L13) script: 

```bash
RUN_KUBEFWD=$(echo  "$RUN_KUBEFWD" | sed 's/ /\\ /g')
DIR=$(echo  "$DIR" | sed 's/ /\\ /g')
```

but then get a :

```
+ /usr/bin/osascript -e 'do shell script "/Users/eric/Library/Application\ Support/tilt-dev/tilt_modules/github.com/tilt-dev/tilt-extensions/kubefwd/run-kubefwd.sh /opt/homebrew/bin/kubefwd /Users/eric/.kube/config /opt/homebrew/bin/entr /var/folders/xh/003_hq_j3qq29mtv3nn8kw6c0000gn/T//tilt-kubefwd.nXOFvvLqd" with administrator privileges'
./sudo-kubefwd.sh: line 43: /Users/eric/Library/Application\ Support/tilt-dev/tilt_modules/github.com/tilt-dev/tilt-extensions/kubefwd/watch-namespaces.sh: No such file or directory
49:50: syntax error: Expected “"” but found unknown token. (-2741)
```


```
# tilt doctor                                                                                                                                                                                                                                   
Tilt: v0.26.3, built 2022-03-24
System: darwin-arm64
---
Docker
- Host: [default]
[...]

PR #381, opened 2022-04-07
kubefwd: copy `run_kubefwd` scripts into a new directory to avoid path normalization issues

lizzthabet:
Based on @nicks' suggestion for #368, this PR follows the pattern in the [`syncback` extension](https://github.com/tilt-dev/tilt-extensions/blob/master/syncback/krsync.sh#L15-L21), which copies scripts to a `tmp` directory before running them.

I'm on MacOS, so I tested these changes by moving the extension `Tiltfile` into a parent directory with a space in its name. Without the changes, the kubefwd scripts fail and with them, they run successfully.

I'm not very familiar with this extension, so feel free to chime in with any technical / style suggestions!


nicks (approved):
nice! LGTM

PR #388, opened 2022-04-20
helm_resource: Add support for charts that use three fields for passing image references

oconnormi:
Currently the `helm_resource` extension supports passing image references to charts either as a single string or as a tuple that maps to two fields in a values file (the default created by `helm create`). This PR adds support for an additional type of structure that uses three fields in a values file:

```yaml
image:
  registry: docker.io
  repository: foo/bar
  tag: 1.0.0
```

A second chart has been added to test this scenario and the test `Tiltfile` has been updated to deploy both simultaneously so as to ensure backwards compatibility with the previous tuple functionality.

### DIFF
--- a/helm_resource/helm-apply-helper.py
+++ b/helm_resource/helm-apply-helper.py
@@ -7,6 +7,15 @@ import os
 import re
 import subprocess
 import sys
+from typing import Dict
+
+def _parse_image_string(image: str) -> Dict:
+  if '.' in image or 'localhost' in image or image.count(":") > 1:
+    registry, repository = image.split('/', 1)
+    repository, tag = repository.rsplit(':', 1)
+    return {"registry": registry, "repository": repository, "tag": tag}
+  repository, tag = image.rsplit(':', 1)
+  return {"registry": None, "repository": repository, "tag": tag}
 
 flags = sys.argv[1:]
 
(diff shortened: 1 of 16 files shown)


helm_resource/helm-apply-helper.py
```diff
@@ -18,13 +18,17 @@
...
 
+  key0 = os.environ.get('TILT_IMAGE_KEY_REGISTRY_%s' % i, '')
   key1 = os.environ.get('TILT_IMAGE_KEY_REPO_%s' % i, '')
   key2 = os.environ.get('TILT_IMAGE_KEY_TAG_%s' % i, '')
-  parts = image.split(':')
-  repo = ':'.join(parts[:len(parts)-1])
-  tag = ':'.join(parts[len(parts)-1:])
-  flags.extend(['--set', '%s=%s' % (key1, repo),
-                '--set', '%s=%s' % (key2, tag)])
+  registry, image = image.split('/', 1)
```
milas, line 24: This won't work if there is no registry, which is a common setup for e.g. Docker Desktop or some minikube configurations:

```
   helloworld │ STEP 3/3 — Deploying
   helloworld │      Running cmd: python3 /Users/milas/dev/tilt-extensions/helm_resource/helm-apply-helper.py --create-namespace
   helloworld │      Traceback (most recent call last):
   helloworld │        File "/Users/milas/dev/tilt-extensions/helm_resource/helm-apply-helper.py", line 24, in <module>
   helloworld │          registry, image = image.split('/', 1)
   helloworld │      ValueError: not enough values to unpack (expected 2, got 1)
```

If you want to test locally w/ minikube:
```sh
minikube start --driver=docker --container-runtime=docker
./helm_resource/test/test.sh
```

milas (changes requested):
This looks good, thanks!

helm_resource/helm-apply-helper.py
oconnormi: Yep, I figured this might be a problem for certain use-cases but hadn't run into it myself (as I usually use k3d). I'll look at how to handle this as well

oconnormi: Should be all set now, I tested it on k3d and docker-desktop

milas (approved):
Thanks for the PR!

PR #390, opened 2022-04-25
k8s_attach: namespace arg works

thhermansen:
As list.append https://github.com/bazelbuild/starlark/blob/master/spec.md#list%C2%B7append
seem to only take 1 arg.


nicks (approved):
thanks! LGTM

PR #392, opened 2022-04-28
Hotfix/make fallback on docker build work

Konubinix:
In some work project, I've realized that the fallback on docker_build does not work, because tilt does not accept None as meaning not given.

I then have made the call to docker_build only pass the arguments that are not None.

I don't know how to update the test to test this change, because:
1. it is not possible to enable image_build for one tilt resource and disable it for the other
2. the structure of the test suggests that we should have only one Tiltfile

Anyway, with another test/Tiltfile with the following content, we can reproduce the issue.

```
load('../Tiltfile', 'image_build', 'kubectl_build_enable')

k8s_yaml('deployment.yaml')
kubectl_build_enable(False)
image_build('example-html-image', '.')
```


nicks (approved):
thanks @Konubinix ! I'm also going to file a bug upstream for this

Konubinix:
Great. Thank you :-).

PR #396, opened 2022-05-09
helm_resource: only modify metadata

nicksieger:
Extract metadata: and trailing indented lines, and only perform namespace
insertion on that text. This was causing some charts to fail to deploy, for
example if a chart has a CRD with an openapi spec that had a `namespace:` key,
the key was erroneously removed by the previous apply helper.

### DIFF
--- a/helm_resource/helm-apply-helper.py
+++ b/helm_resource/helm-apply-helper.py
@@ -74,22 +74,29 @@ def add_default_namespace(yaml, namespace):
   if not namespace:
     return yaml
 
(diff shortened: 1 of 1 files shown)


nicksieger:
More namespace default helper bugz ☹️

nicks:
can you add this case to the unit test? maybe add a ConfigMap with a 'namespace' key and assert the key is perserved?

also the unit test appears to be failing.

nicks (approved):
woooo

PR #398, opened 2022-05-17
kubectl_build: fix docker_build empty default values

glehmann:


glehmann:
Sorry, I forgot to fill the description – it's not good to do things in a hurry!

It fixes this error when building an image with the default `secret` value:

~~~
Build Failed: ImageBuild: could not parse secret: []: failed to parse csv secret: EOF
~~~

It was working before, I'm not sure exactly what changed — maybe the switch to a newer tilt version

nicks (approved):
thanks!

PR #400, opened 2022-05-26
earthly: build extension

konradmalik:
Hey all,
I've created an extension that allows to use [Earthly](https://docs.earthly.dev/) as a tool to build images for Tilt.

The code is largely based on the `podman` extension.

Let me know if something needs changing.


konradmalik:
I added earthly binary to the dev Dockerfile (needed by this extension) but I'm not sure how to update the ref so that the tests could use it.

nicks:
ya, let me add this to the ci image

nicks:
I merged this independently as https://github.com/tilt-dev/tilt-extensions/pull/410

PR #401, opened 2022-05-26
Pack: fix docker tag on windows

### LINKED ISSUE #294: Pack extension fails when running on Windows
Pack extension fails when tilt is run on Windows with `Error parsing reference: "$EXPECTED_REF" is not a valid repository/tag: invalid reference format: repository name must be lowercase`. Same tiltfile works fine when run under WSL (Ubuntu).

Tiltfile:
```python
allow_k8s_contexts('gke_fe-astakhov_us-central1-c_gke-play')

load('ext://pack', 'pack')

pack('macsux/dotnet-accelerator', 
    env_vars=['BP_LIVE_RELOAD_ENABLED=true', 'BP_DOTNET_PROJECT_PATH=./src/MyProjectGroup.DotnetAccelerator'])

k8s_yaml(['kubernetes/deployment.yaml'])
k8s_resource('dotnet-accelerator', port_forwards=8080)
```

iainsproat:
Resolves #294

### DIFF
--- a/pack/Tiltfile
+++ b/pack/Tiltfile
@@ -44,10 +44,12 @@ def pack(
       " ".join(["--env " + s for s in envs]),
     ])
 
+    isWindows = True if os.name == "nt" else False
+    expected_ref = "%EXPECTED_REF%" if isWindows else "$EXPECTED_REF"
     docker_tag_cmd = " ".join([
       "docker tag",
       caching_ref,
-      "$EXPECTED_REF",
+      expected_ref,
     ])
 
     custom_build(


nicksieger:
Thanks for the contribution! Does this fix the extension for you? The `custom_build` command has an `&&` in it which I wouldn't think works on Windows either. We might want to instead use the `command_bat=` parameter for Windows rather than the conditional you have here. What do you think?

iainsproat:
Hi @nicksieger - yes, thanks for reviewing this. The changes are successfully running on my windows 10 machine (see below screenshot).

I've just tested that the `&&` works in the windows command prompt - it worked for me.
image

nicksieger:
Thanks for confirming! TIL!

PR #411, opened 2022-06-14
feat(ngrok): Ngrok extension to support V3

adinhodovic:
https://ngrok.com/docs/guides/upgrade-v2-v3

Do I need to add compability for Ngrok v2 agent (v1 config)?

Also added a flag for specifying default_ngrok_config.

### DIFF
--- a/ngrok/README.md
+++ b/ngrok/README.md
@@ -7,7 +7,7 @@ Create public URLs for your local services with [`ngrok`](https://ngrok.com/)
 ## Requirements
 
 - `bash`
-- `ngrok v3`
+- `ngrok`
 - GNU core utils (`tr`, `sort`) - `brew install coreutils`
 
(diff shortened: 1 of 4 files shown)


nicks:
ya, i think we probably need a way for people to transition? i'd be fine with a `version` param, and use the old config on version 2

adinhodovic:
> ya, i think we probably need a way for people to transition? i'd be fine with a `version` param, and use the old config on version 2

Would that work?

JakulaTechuniverse:
Can someone please help me install ngrok agent v3 ?

JakulaTechuniverse:
I managed . I guess they no longer support phishing . Coz i was trying to demonstrate phishing with blackeye to my students and they blocked me . I guess i'll have to find another ngrok alternative for phishing .

Issue #374, opened 2022-04-04
helm_resource fails if chart install resources in other namespaces


bamorim:
If I do, for example:

```starlark
helm_repo("prometheus-community", "https://prometheus-community.github.io/helm-charts")

helm_resource(
    "kube-prometheus-stack",
    "prometheus-community/kube-prometheus-stack",
    namespace = "o11y",
    flags = [
        '--version', '34.8.0',
        '--values', os.path.join(this_dir, "kube-prometheus-stack.yaml")
    ]
)
```

This will fail because `kube-prometheus-stack` install some things in the namespace `kube-system`, so `helm get manifest --namespace o11y kube-prometheus-stack | kubectl get --namespace o11y -oyaml -f -` fails with:

```
the namespace from the provided object "kube-system" does not match the namespace "o11y". You must pass '--namespace=kube-system' to perform this operation.
the namespace from the provided object "kube-system" does not match the namespace "o11y". You must pass '--namespace=kube-system' to perform this operation.
the namespace from the provided object "kube-system" does not match the namespace "o11y". You must pass '--namespace=kube-system' to perform this operation.
the namespace from the provided object "kube-system" does not match the namespace "o11y". You must pass '--namespace=kube-system' to perform this operation.
the namespace from the provided object "kube-system" does not match the namespace "o11y". You must pass '--namespace=kube-system' to perform this operation.
```

I don't know what the best approach would be. Maybe we should get all manifests  with defined namespaces and get from that namespace instead?

nicks:
I investigated this a bit. There's a bunch of overlapping problems with this chart that make this difficult to fix. Basically:

1) This chart has unusually complex namespacing semantics. Some resources explicitly specify kube-system. Some resources inherit the namespace from the chart namespace. Some resources inherit the namespace from the default kubecontext (see: https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/servicemonitor.yaml#L8)

2) In particular, Helm chart namespace and Helm default kubecontext namespace are slightly different

3) `helm get manifest` fetches all the resources in the chart. Helm chart namespace is filled in. Helm default kubecontext namespace is not filled in.

4) `kubectl get --namespace [namespace] -oyaml -f -` uses the `--namespace` flag as both "default kubecontext namespace" and as "please validate all the returned resources match this namespace"

I could probably make the case that the `kubectl` behavior is incorrect here. The `You must pass '--namespace=%s'` error probably isn't intended for this use-case...

[...]

PR #413, opened 2022-06-27
helm_resource: fix helm_resource() so it will deploy to custom namespaces

mars-f:
Make `helm_resource()` pass its namespace parameter to the `kubectl` command, not just the `helm` command.

This fixes a bug where deploying a new service into a custom namespace would fail on the `kubectl get -oyaml -` step.  `helm_resource()` calls to Helm use `helm --namespace SOME_NAME` to work on a specific namespace. `helm_resource()` calls to `kubectl` would look in the default namespace for the resources that Helm just created in the custom namespace, causing `kubectl` to return an error code.

### DIFF
--- a/helm_resource/helm-apply-helper.py
+++ b/helm_resource/helm-apply-helper.py
@@ -57,6 +57,7 @@ namespace = os.environ.get('NAMESPACE', '')
 if namespace:
   install_cmd.extend(['--namespace', namespace])
   get_cmd.extend(['--namespace', namespace])
+  kubectl_cmd.extend(['--namespace', namespace])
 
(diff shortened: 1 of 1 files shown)


nicksieger (approved):
Thanks for the PR! Looks good to me.

nicksieger:
Sorry @mars-f. I spoke too soon. See https://github.com/tilt-dev/tilt-extensions/issues/374#issuecomment-1089271912 for details.

nicks:
can you file a separate bug for the problem you're seeing? namespace defaulting is very complex and subtle, which is why we have tons of tests for this.

PR #416, opened 2022-06-27
test: remove pulumi secrets from circleci

nicks:
Hello @nicksieger,

Please review the following commits I made in branch nicks/pulumi4:

293894bbe7ac85df9c6c5e71fb897c260a81e085 (2022-06-27 18:23:56 -0400)
test: remove pulumi secrets from circleci

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,6 +36,4 @@ workflows:
     jobs:
       - validate
       - shellcheck
-      - test:
-          context:
-            - Tilt Pulumi Context
+      - test


nicks:
circleci keeps changing the rules around how secrets are injected into tests when there are external PRs. Looks like their most recent change is "just don't run the tests".

the best solution would be to separate out tests that need secrets with tests that don't need secrets, but doesn't seem really worth the effort right now.

nicksieger (approved):

PR #417, opened 2022-06-29
earthly: add windows-specific command and restarts

konradmalik:
Turns out that for windows-native earthly installations the `$EXPECTED_REF` variable was not expanded properly.
Needed to add a windows-specific command via `command_bat`.

Additionally, I added a "docker_build_with_restart"-like functionality.


nicksieger (approved):
Looks great! Thanks!

PR #418, opened 2022-07-01
Nix flake extension

graysonhead:
A heavily modified version of [afreakk](https://github.com/afreakk)'s Nix plugin that allows it to build container Derivations from a Nix flake.


nicks:
Nice! 

We'll need to exclude this test from CI for now, until we figure out how to get Nix working correctly in a circleci container. Can you add this test to the list here:
https://github.com/tilt-dev/tilt-extensions/blob/master/test.py#L33

(until then we'll just run the test manually)

PR #428, opened 2022-07-26
Expose disabled and requires_confirmation for uibutton extension

MarcDufresne:
Hi all,

We've encountered use cases for having disabled buttons, as well as buttons that require confirmation (e.g. destructive actions), and we noticed that those features are supported by Tilt. However, the uibutton extension doesn't expose those params. We tested those changes in our Tiltfile and felt like it would be a quick win to share this to the main repo.

Let me know what you think!


nicks (approved):
thanks! LGTM

PR #434, opened 2022-08-10
Add support for kubectl_build `--push` argument

jbeard6:
# What is changed

Adding `push` as an optional parameter to `kubectl_build` that appends the `--push` argument to the `kubectl build` command line. This defaults to `False`, maintaining the existing behavior.

# Why

When using an external registry with Tilt, the images built by `kubectl_build` must be pushed to the registry in order to be pulled and started by tilt. This is handled by the `kubectl build` command, but needs to be exposed to the corresponding tilt function.


nicks (approved):
LGTM! thanks

Issue #435, opened 2022-08-10
helm-resource: different namespaces breaking on Windows


markmixson:
I'm installing an oracle db helm chart (https://github.com/oracle/docker-images/tree/main/OracleDatabase/SingleInstance/helm-charts/oracle-db) using helm_resource in a separate namespace called "oracle" using Minikube.  On Macs/Linux, this works fine.  But on Windows, the helm-apply-helper.py script gets confused.

```
     Running cmd: ['helm', 'get', 'manifest', '--namespace', 'oracle', 'db19c']
     Running cmd: ['kubectl', 'get', '-oyaml', '-f', '-']
     Error from server (NotFound): secrets "db19c-oracle-db" not found
     Error from server (NotFound): persistentvolumeclaims "db19c-oracle-db" not found
     Error from server (NotFound): services "db19c-oracle-db" not found
     Error from server (NotFound): deployments.apps "db19c-oracle-db" not found
     Traceback (most recent call last):
       File "C:\Users\ss4ro\AppData\Local\tilt-dev\tilt_modules\github.com\tilt-dev\tilt-extensions\helm_resource\helm-apply-helper.py", line 77, in <module>
         completed.check_returncode()
       File "C:\Users\ss4ro\AppData\Local\Programs\Python\Python38\lib\subprocess.py", line 444, in check_returncode
         raise CalledProcessError(self.returncode, self.args, self.stdout,
     subprocess.CalledProcessError: Command '['kubectl', 'get', '-oyaml', '-f', '-']' returned non-zero exit status 1.

Build Failed: apply command exited with status 1
```

When I change helm-apply-helper.py to include the namespace with kubectl_cmd, everything works fine:
```
if namespace:
  install_cmd.extend(['--namespace', namespace])
  get_cmd.extend(['--namespace', namespace])
  kubectl_cmd.extend(['--namespace', namespace])
```

I have not tried to figure out why this works on Mac/Linux but not Windows though. . .

markmixson:
I tried to fix this trivially in #437 but appears that's wrong.  Also the tests for helm_resource do not pass in Windows.

helm 3.9.2
kubectl 1.24.3
minikube 1.26.1
kubernetes 1.24.3
tilt 0.30.6

```
++ dirname '.\test.sh'
+ cd .
+ tilt ci
Tilt started on http://localhost:10350/
v0.30.5, built 2022-07-14

Initial Build
[...]

PR #437, opened 2022-08-10
helm_resource: add namespace to kubectl command

markmixson:
Signed-off-by: Mark Mixson <mixson_mark@bah.com>

### DIFF
--- a/helm_resource/helm-apply-helper.py
+++ b/helm_resource/helm-apply-helper.py
@@ -58,6 +58,7 @@ namespace = os.environ.get('NAMESPACE', '')
 if namespace:
   install_cmd.extend(['--namespace', namespace])
   get_cmd.extend(['--namespace', namespace])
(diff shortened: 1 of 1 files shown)


markmixson:
```
   helloworld │      error: the namespace from the provided object "kube-public" does not match the namespace "app". You must pass '--namespace=kube-public' to perform this operation.
```
Not understanding how "kube-public" is involved here.

markmixson:
references #435

nicks:
you're not the first person who has made this mistake, that's why we have tests for it :sweat_smile: 

kubectl namespace defaulting and helm namespace defaulting have different semantics. Even though the flags are both called `--namespace`, they do different things.

markmixson:
I'm trying to fix something that is broken.  Based on #435 I had to do this to get the helm chart to work in Windows 🤷

markmixson:
Apparently there's bigger problems than this in Windows. . . .

PR #438, opened 2022-08-11
Adds Windows support to helm_resource (#435)

### LINKED ISSUE #435: helm-resource: different namespaces breaking on Windows

markmixson:
Signed-off-by: Mark Mixson <mixson_mark@bah.com>

fixes #435

### DIFF
--- a/helm_resource/Tiltfile
+++ b/helm_resource/Tiltfile
@@ -69,9 +69,11 @@ def helm_resource(
 
   if len(image_deps) != len(image_keys):
     fail('image_deps and image_keys must have the same length, so that we know how to inject images into the helm chart;')
-
-  apply_cmd = ['python3', _helm_apply_path]
-  delete_cmd = ['python3', _helm_delete_path]
+  
+  is_windows = True if os.name == "nt" else False
+  python_cmd = ['py', '-3'] if is_windows else ['python3']
+  apply_cmd = python_cmd + [_helm_apply_path]
+  delete_cmd = python_cmd + [_helm_delete_path]
   env = {
     'CHART': chart,
     'RELEASE_NAME': release_name,
--- a/helm_resource/helm-apply-helper.py
+++ b/helm_resource/helm-apply-helper.py
@@ -4,7 +4,6 @@
 # python3 helm-apply-helper.py ... [image config keys in order]
 
 import os
-import re
 import subprocess
 import sys
 from typing import Dict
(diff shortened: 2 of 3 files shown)


nicks (approved):
Nice change! LGTM

markmixson:
As a follow up to this issue, I found the real culprit for why I had to do this.

The Oracle helm chart is not available in an official helm repository, so as part of a Tilt build we use, we pull a git repo and install the chart that way.  When we pull the git repo in Windows, it sticks Window line endings throughout the Helm chart.  Most tools handle that OK, but `helm_resource` needed to convert them to do the default namespacing bit.  It's also unpredictable if a Helm chart will have Windows line endings or not.

As an optimization, it might be better to use the `os.linesep` instead of `\n` in the default namespacing code to avoid having to filter the kubectl output like this.

PR #442, opened 2022-08-16
Additional kubectl_build options

jbeard6:
# What is changed

Add kubectl_build builder options `namespace` and `builder`, corresponding to `kubectl-build` command line options `--namespace` and `--builder`. These options are used by `kubectl-build` to locate the builder pod to be used.

# Why

These options allow the builder pod to be located in a common namespace colocated with the registry secret.


nicks (approved):
LGTM!

Issue #384, opened 2022-04-13
Use builtin for uibutton extension


lianmakesthings:
On Windows, adding `cmd_button`s to the Tiltfile causes evaluation to take several seconds longer per button.

```
Trying to use cmd_button but it seems like each use of it adds several seconds to the tiltfile parsing. Is this supposed to happen? It wouldn't be too bad if it was just on initial load but it happens on any changes to any resources too.
Looks like the apply it calls is very slow eg:
verb="PATCH" URI="/apis/tilt.dev/v1alpha1/cmds/btn-button1?fieldManager=kubectl-client-side-apply" latency="6.7736ms" userAgent="Go-http-client/2.0"
```
```
load('ext://uibutton', 'cmd_button', 'location')
cmd = ['hostname']

local_resource('local', cmd='hostname')

cmd_button('button1',argv=cmd,location=location.NAV)
cmd_button('button2',argv=cmd,location=location.NAV)
cmd_button('button3',argv=cmd,location=location.NAV)
cmd_button('button4',argv=cmd,location=location.NAV)
```
```
> tilt up --stream --port 10351
Tilt started on http://localhost:10351/
v0.27.0, built 2022-04-01

Initial Build
Loading Tiltfile at: C:\temp\tiltdev\Tiltfile
 → uibutton.tilt.dev/button1 created
 → cmd.tilt.dev/btn-button1 created
 → uibutton.tilt.dev/button2 created
 → cmd.tilt.dev/btn-button2 created
 → uibutton.tilt.dev/button3 created
 → cmd.tilt.dev/btn-button3 created
 → uibutton.tilt.dev/button4 created
 → cmd.tilt.dev/btn-button4 created
Successfully loaded Tiltfile (45.6043661s)
```

We are currently unable to reproduce this behaviour on our local machines
Currently, the `cmd_button` extension shells out to `tilt appl`y which is probably causing unnecessary overhead.
Our best guess is, that using the builtin instead of shelling should cut down on processes spawned.

lianmakesthings:
Another user reports the same issue and their workaround

```
Hey, I was wondering what the impact would be of removing all the local() invocations in https://github.com/tilt-dev/tilt-extensions/blob/master/uibutton/Tiltfile and hardcoding the "startAfter": _now() to a fixed timestamp.
My problem is that the local invocations are slow on my windows machine, adding 1 button adds +- 2 seconds loading time to the tiltfile because the local command starts powershell and this process is slow.
[...]

Issue #332, opened 2022-02-08
ui_button behaves badly with duplicate button names


landism:
If a Tiltfile creates multiple buttons with the same name, only the last one will be kept. No error will be generated.

This can be solved by migrating the ui_button extension to use `v1alpha1.ui_button` builtins, which were added in v0.22.10.

(that migration will also fix #193)

Issue #193, opened 2021-06-29
cmd_button does not remove old/renamed buttons


milas:
If you have `cmd_button('foo', resource='bar', cmd=['echo', 'hi'])` and either remove it from `Tiltfile` or change the button ID (`foo` -> `foo2`), the old button will not be removed.

The workaround for now is to restart Tilt as buttons are not persistent across `tilt up` runs.

milas:
I think this is an unlikely scenario, but if you have a button whose text is dynamic, you can also use the optional `text` argument and provide a stable value as the first argument to be used as the ID:

```python
cmd_button(
  'this-should-not-change',
  resource='my-resource',
  cmd=['echo', 'hi'],
  text=_dynamically_generate_button_text()
)
```

PR #443, opened 2022-08-18
Use builtins for uibutton extension

jbadlato:
Fixes: #384 #193 #332 

I smoke tested the case from #193 to confirm that renamed buttons now get cleaned up properly


nicks (approved):
oops, i thought i had merged this! LGTM

PR #447, opened 2022-10-10
Rename the binary directory not to be ignored

LittleWat:
kubebuilder currently ignores `bin/` by dockerignore https://github.com/kubernetes-sigs/kubebuilder/blob/2ba91dfb7c4f45b3ee47ebed07efb0cfbbe9abcf/pkg/plugins/golang/v3/scaffolds/internal/templates/dockerignore.go#L43

This causeed the error (`bin` not found error) so I would like to modify the directory name.


LittleWat:
Thank you for your review and for adding the comment!

PR #449, opened 2022-10-14
feat: Add flag to configure tls scheme used by tunnels

Xaelias:
Adding `--ngrok_config_tls_scheme` as an arg with a value of `https` will create https tunnels instead of http ones.


nicks (approved):
thanks! LGTM

PR #459, opened 2022-11-26
execute_in_pod: run a command inside of a pod container

Llandy3d:
Hi, this extension allows the user to specify a command that will be run inside of a pod container.
The extension does the heavy work of figuring out on which pod/container it should run.
Currently it will only run the command on a single container.


nicks:
i'm a little bit confused how this works? if you execute it in a `local()`, the kubernetes discovery objects won't exist yet. 

can you add a test with an example usage? here's a good small test you can borrow from: https://github.com/tilt-dev/tilt-extensions/tree/master/namespace/test

Llandy3d:
Having a running pod the idea is that you might want to run some commands inside of it. For example you might want to run some things to configure your application or similar and those commands have to be run inside of the specific container.

To do that you can create a `local_resource` that depends on a k8s resource and make it run the command given by the function. A practical example would be creating a django superuser and/or loading fixtures and you need those operations to be run from the container.

I've been using it with the mode that returns the command to be given to `local_resource` but I've added an execute mode that now I can see that maybe won't be useful/work 🤔 

But yeah the gist of it was to be able to run a command inside of a container programmatically, when not defined in yaml files 😄

nicks:
maybe the api should be something like:
execute_in_pod_cmd - returns the command array
execute_in_pod_resource - creates a local resource

sort of like how we set up the deployment extension to have versions that return yaml or set up deployers (https://github.com/tilt-dev/tilt-extensions/tree/master/deployment)

Llandy3d:
I've removed the execute option and fixed the examples on how to use the extension, for now only with `local_resource` as the command will return the command to run.

I wasn't able to try to create the tests now and with the vacations coming up it might slip, if it's not enough I understand!
The important bit is that if someone wants/needs to do the same could see the pr and take inspiration 😁
I wish you great festivities!

nicks (approved):
LGTM!

PR #465, opened 2022-12-14
dotenv: do not fail if file is missing

sevein:
I've found that dotenv fails if the env file is missing because it uses read_file without a default.

Let me know if this is something that you'd consider!

### DIFF
--- a/dotenv/Tiltfile
+++ b/dotenv/Tiltfile
@@ -1,5 +1,8 @@
-def dotenv(fn='.env'):
-	f = read_file(fn)
+def dotenv(fn='.env', required=True):
+	default = None
+	if not required:
+		default = ""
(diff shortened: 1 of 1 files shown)


nicks:
Hmmmm...I'm not sure. 

I poked at other tools that process env files (e.g., Docker Compose), they emitted an error if the env file doesn't exist.

Can you give more explanation for why you think the API should behave this way?

milas:
You can also wrap invocation based on an existence check to achieve this behavior:
```python
dot_env_path = '.env'
if os.path.exists(dot_env_path):
   dotenv(dot_env_path)
```

sevein:
Agree, it doesn't look like it's necessary if I can simply use os.path.exists.

Thank you for the tip! :heart_eyes:

PR #468, opened 2023-01-11
namespace: add support for ServiceAccount subjects in rolebindings an…

Cyanopus:
…d clusterrolebindings

This PR is for extending the namespace inject to work with ClusterRoleBindings and RoleBindings. An example of such manifests is the ArgoCD installation. Adding this support will allow for the proper apply of the manifest via `k8s_yaml`.


nicks (approved):
seems reasonable to me! i can't think of a real use-case where you'd want to run a ClusterRole thru namespace_inject but not update the namespaces of the subjects.

PR #471, opened 2023-01-20
helm_resource: Add `auto_init` keyword argument

noelleleigh:
Include support for the `auto_init` keyword argument from [`k8s_resource`][0].

The use-case is our team wants to be able to run a service from a Helm chart in our cluster, but it is not critical to the functionality of the cluster and incurs a significant performance cost. We would like to be able opt-out of starting this heavy service automatically when running `tilt up`.

No tests are included because I'm not certain how to test that a resource _doesn't_ exist in a meaningful way.

[0]: https://docs.tilt.dev/api.html#api.k8s_resource

### DIFF
--- a/helm_resource/README.md
+++ b/helm_resource/README.md
@@ -32,7 +32,8 @@ def helm_resource(
     live_update=None,
     resource_deps=None,
     labels=None,
-    port_forwards=[]):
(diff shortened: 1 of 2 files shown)


nicks (approved):
LGTM!

PR #475, opened 2023-01-25
vault_client: reaching secrets from a external vault

ismferd:


ismferd:
@nicksieger Can you help me merging that (if it is possible)?

nicks:
seems reasonable to me.

we'll need you to add a signed-off field before we can accept this contribution - https://github.com/tilt-dev/tilt-extensions/pull/475/checks?check_run_id=10870989216

can you also add this to the list of extensions in the root README as part of this PR? https://github.com/tilt-dev/tilt-extensions/blob/master/README.md

nicks (approved):
LGTM!

PR #480, opened 2023-02-23
helm_resource: Make helm-delete-helper quieter

noelleleigh:
This PR makes `helm-delete-helper` smarter by checking if a release is installed _before_ trying to uninstall it.

After #471 was merged, it's possible for Helm releases to be known to Tilt, but not actually running in the cluster. This means that when `tilt down` is called, the [`helm uninstall`][0] command run by `helm-delete-helper` will fail with an error like:

> Error: uninstall: Release not loaded: kafka-operator: release: not found

While these errors are harmless, they are noisy and can appear as if Tilt itself is failing. This PR aims to remove those errors altogether by calling [`helm status`][1] first and checking that it was successful before calling `helm uninstall`.

The rest of the code was also refactored to be simpler.

To test this, run the `/helm_resource/tests/test.sh`. When it tears down the resources it created, this error should not appear:

> Error: uninstall: Release not loaded: memcached-no-init: release: not found

[0]: https://helm.sh/docs/helm/helm_uninstall/
[1]: https://helm.sh/docs/helm/helm_status/

### DIFF
--- a/helm_resource/helm-delete-helper.py
+++ b/helm_resource/helm-delete-helper.py
@@ -7,13 +7,18 @@ import os
 import subprocess
 import sys
 
-delete_cmd = ['helm', 'uninstall']
+release_name = os.environ['RELEASE_NAME']
 namespace = os.environ.get('NAMESPACE', '')
-if namespace:
-  delete_cmd.extend(['--namespace', namespace])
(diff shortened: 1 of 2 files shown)


nicks (approved):
LGTM!

nicks:
thanks!

PR #483, opened 2023-03-06
get_obj: add new extension

kobik:


kobik:
Hi @nicks , I'd be happy if you could review this.

Cheers.

remote_development/Tiltfile
```diff
@@ -0,0 +1,82 @@
...
+    fail("failed to find container %s in the manifest %s" % (container_selector, containers))
+
+def get_registry(container):
+  return container["image"].rsplit(":", 1)[0]
+
+def remove_probe(container, probe):
+  if container.get(probe):
+    container.pop(probe)  
+
+def expose_ports(container, port_forwards):
```
nicks, line 18: i don't think you need this?

From the docs: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#container-v1-core

> List of ports to expose from the container. Exposing a port here gives the system additional information about the network connections a container uses, but **is primarily informational.** [emphasis added]

remote_development/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+Author: [Kobi Carmeli](https://github.com/kobik)
+
+Develop, live update and debug against applications that were already deployed to Kubernetes from CI.
+
+### `remote_development(obj: str, container_selector: str, namespace: str, port_forwards: port_forwards, disable_probes: bool, live_update: live_update, dockerfile: str, context: str)`
+- **obj**: An object name in the form "kind/name", as you would express it to kubectl. e.g. deployment/my-deployment.
+- **container_selector**: The container name in the pod that will be used for remote development.
+- **namespace**?: The namespace of the object. If not specified, uses the current namespace.
+- **dockerfile**?: The dockerfile that would be used for building a development image suitable for live-update.
```
nicks, line 11: i think my main concern here is that docker_build and k8s_resource take a lot of arguments, so you're going to need to keep adding more args here.

what if there was a helper function? then your helper function could return something like:

```
def remote_get(obj):
  ...
  return struct(yaml=yaml, image=image)
 ```

and then people who wanted to use different args could use `remote_get` instead of `remote_development`

Comment (kobik) [reply remote_development/README.md:11]: I get your point and it was also my concern, I made it simple for a simple usage for simple cases, while keeping `dockerfile` and `port_forwards` optional, so users could cover various cases with their own `docker_build` and `k8s_resource` and so, it could be used like that in these cases:
```python
name='example-nodejs'
docker_build(image, '.', # image is now something the users should set by their own
  dockerfile='Dockerfile.dev',
  live_update=[
    sync('./app', '/app')
  ]
)
remote_development(
  obj='deployment-{}'.format(name),
  container_selector='my-container',
  remove_probes=True
)
k8s_resource(
  name,
  port_forwards=[
    "8080:8080",
    "9229:9229"
  ]
)
```

If we'll take your suggestion, usage would look something like this
```python
name='example-nodejs'
(yaml, image)=remote_get(
  obj='deployment-{}'.format(name),
  container_selector='my-container', # optional, will also make returned image optional, but would be required for removing probes
  remove_probes=True # optional
)

docker_build(image, '.',
  dockerfile='Dockerfile.dev',
  live_update=[
    sync('./app', '/app')
  ]
)

k8s_yaml(yaml)

k8s_resource(
  name,
  port_forwards=[
    "8080:8080",
    "9229:9229"
  ]
)
```

What would be your go-to option?

I don't mind moving forward with the second option, but we'll need to figure out a better describing name 🙂

remote_development/Tiltfile
kobik: Good to know

remote_development/README.md
kobik: btw, instead of `obj` we can use `name` and `kind` in both cases.

wdyt?

kobik: @nicks i've changed to be `get_obj` which returns a struct of `image`, `registry` and `yaml`.

let me know what you think.

Thanks

nicks (approved):
LGTM!

PR #486, opened 2023-03-13
kubebuilder: disable security context

hrenard:
The default kubebuilder boilerplate includes a security context forbidding root user which is required by Tilt.

This PR adds a `DISABLE_SECURITY_CONTEXT` option, enabled by default to patch the deployment and remove the security context as [this comment](https://github.com/tilt-dev/tilt/issues/3060#issuecomment-779726111).

@jazzdan


nicks (approved):
LGTM

PR #487, opened 2023-03-13
helm_resource: version

kplimack:
* adds ability to provide `--version` argument to `helm_resource`

### DIFF
--- a/helm_resource/README.md
+++ b/helm_resource/README.md
@@ -23,6 +23,7 @@ def helm_resource(
     chart,
     deps=[],
     release_name='',
+    version=None,
     namespace=None,
     image_deps=None,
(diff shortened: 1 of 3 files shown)


jplimack:
@nicks perhaps, but something more declarative is easier to consume and understand the behavior of; hence me not knowing of this way to provide this argument

nicks:
I would be fine with a note on the readme on how to set `--version`. there's already a note about `--set` where you can put it.

I don't think we want to replicate every helm install flag as starlark arguments

nicks:
i added some notes to the README and test on how to use `--version`
https://github.com/tilt-dev/tilt-extensions/tree/master/helm_resource#helm_resource

PR #489, opened 2023-03-16
git_resource: add no-single-branch flag for shallow clones

bkero:
Currently if you reference a repo with a branch,then change that branch it will fail with an error of 'No such reference'. This is because since version 1.8, Git shallow clones only pull a single branch [1]. There was a flag added (--no-single-branch) to pull all branches.

[1] https://stackoverflow.com/questions/23708231/git-shallow-clone-clone-depth-misses-remote-branches


nicks (approved):
LGTM!

PR #492, opened 2023-04-18
kubectl_build: Add support for --platform argument

NoiSek:
This PR adds support for the `--platform` argument when running `kubectl_build`, a standard way to specify build arches supported by both the `docker_build` ext and the underlying Docker CLI.

Crucial component to deploying remotely in variable environments (i.e. clusters running on mixed-arch spot instances).


nicks (approved):
LGTM!

PR #496, opened 2023-05-12
added decoding for special characters in git branch names

MathiasAugstein:
We are using branch names containing `#`, hence the missing possibility to encode these characters is a problem for us. 
I propose this change, which lets the user use encoded special characters for the two characters used to split the branch name from the repo name.

Does this need to be documented in the Readme, or can we trust that smart software developers try encoding the characters according to standard URL encoding rules?

### DIFF
--- a/git_resource/README.md
+++ b/git_resource/README.md
@@ -26,7 +26,8 @@ git_resource(resource_name, path_or_repo, dockerfile='Dockerfile', namespace='de
 
 * `resource_name` ( str ) – the name to use for this resource
 * `path_or_repo` ( str ) – either the URL to the remote git repo, or the path to your local checkout.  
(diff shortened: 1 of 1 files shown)


nicks:
thanks for the report! posted https://github.com/tilt-dev/tilt-extensions/pull/497 as an alternative

nicks:
actually, i changed my mind, i like this pr better. can you add a test case for it under ./test?

MathiasAugstein:
Test cases added. Also decided to add one sentence to the Readme. In case you do not like it I can remove the Readme-Commit.

Anyway, I am glad to solve my problem and contribute to the project. My first time pushing code to an open source repo. :)

nicks (approved):
lgtm!

PR #497, opened 2023-05-12
git_resource: handle branch names with #

nicks:
alternative to https://github.com/tilt-dev/tilt-extensions/pull/496


nicks:
actually, i changed my mind, i like the other pr better

Issue #444, opened 2022-09-18
"dotenv" handles quoted values in unexpected behavior


jeffyongtaotang:
In Node, the "dotenv" extension is not handling double/single quoted value in expected behavior.

Env: nodejs
Tilt version: v0.30.7

#### Example .env file
```
KEY='quoted_value'
```

#### Expected result
```
process.env.KEY = "quoted_value"
```

#### Current result
```
process.env.KEY = "'quoted_value'"
```

Quoted values are suggested from the "dotenv" node implementation [README](https://github.com/motdotla/dotenv/blob/b2a316438983148e598ec75a3823cc361fda420b/README.md?plain=1#L59). I think also true for some other tech stack as well.

Suggesting the "dotenv" extension can process the prefix/suffix single & double quotes. e.g.

```python
def dotenv(fn='.env'):
  f = read_file(fn)
  lines = str(f).splitlines()
  for line in lines:
    v = line.split('=', 1)
    if len(v) < 2:
      continue # skip empty lines
    
    v[1] = v[1].removeprefix("\'").removeprefix("\"")
    v[1] = v[1].removesuffix("\'").removesuffix("\"")
    os.putenv(v[0], v[1])
```

rrmistry:
I was able to get past this by just implementing the same function but directly in my local `Tiltfile` using the code below:

```python
# Reference: https://github.com/tilt-dev/tilt-extensions/tree/master/dotenv
def dotenv(fileName=".env"):
    if not os.path.exists(fileName):
        print("❓ Environment variables file not found: %s. Skipping load and continuing..." % fileName)
        return
    print("💾 Loading environment variables from file: %s" % fileName)
    f = read_file(fileName)
    lines = str(f).splitlines()
    for line in lines:
        v = line.split('=', 1)
        if len(v) < 2:
            continue # skip empty lines
        print("⚙️ Loaded environment variable %s" % v[0])
        os.putenv(v[0], v[1].strip().strip('"').strip("'"))

k8s_context_name = k8s_context()
env_file_name = "%s.env" % k8s_context_name
print("☁️ Using Kubernetes context specific file: %s" % env_file_name)
dotenv(fileName=env_file_name)
```

I hope this helps, though it would be good to have it fixed directly in the extension.

Can send a PR soon if time permits.

PR #498, opened 2023-05-16
Enhancing dotenv function

rrmistry:
Enhancing the `dotenv` function with few enhancements:

* Handling environment variable names and values with quotes (`'` or `"`)
* Handling comments in ".env" variable file
* Gracefully moving on if ".env" file name not found
* Printing line, file name, and line number for invalid formatted lines in ".env" files

This should resolve https://github.com/tilt-dev/tilt-extensions/issues/444


dotenv/Tiltfile
```diff
@@ -1,8 +1,46 @@
...
+            if len(line) > 0:
+                fail("Invalid format for dotenv file: '%s' in '%s:%s'" % (line, fn, lineNumber+1))
+            continue # skip empty lines
+
+        # Print if key is valid
+        print("Loading environment variable '%s'" % v[0])
+
+        # Set environment variable
+        os.putenv(
+            v[0].strip().strip('"').strip("'"),
```
nicks, line 44: i think this will take out all the quotes anywhere in the string, which is not what we want?

dotenv/Tiltfile
```diff
@@ -1,8 +1,46 @@
...
+        )
+
+        # Check if key and value are present
+        if len(v) < 2:
+            if len(line) > 0:
+                fail("Invalid format for dotenv file: '%s' in '%s:%s'" % (line, fn, lineNumber+1))
+            continue # skip empty lines
+
+        # Print if key is valid
+        print("Loading environment variable '%s'" % v[0])
```
nicks, line 40: this seems like it'd be very chatty?

dotenv/Tiltfile
```diff
@@ -1,8 +1,46 @@
...
-			continue # skip empty lines
-		os.putenv(v[0], v[1])
+def dotenv(fn=".env"):
+
+    # Check if file exists
+    if not os.path.exists(fn):
+        warn("Environment variables file not found: '%s'" % fn)
+        return
+
+    print("Loading environment variables from file: '%s'" % fn)
```
nicks, line 8: this seems like an odd thing to print?

nicks (commented):
cc @milas , who had a lot of grief over Compose's env file parser. curious what he thinks of this. i realize this extension isn't as widely used as Compose's parser, and so the trade-offs might be different.

dotenv/Tiltfile
rrmistry: The intention here was to provide visibility into which dotenv file is currently being loaded.

Our use case is to have a Git Mono Repo that contains Git Sub-Repos for each of the micro services. And we dynamically load the Titlfile from the sub-repos via the Tiltfile in the Mono Repo.

IMHO, this is still good to have. But if this seems to verbose then I can think of adding a `verbose` or `debug` parameter in the parent function that would provide this diagnostic information.

What are your thoughts?

```suggestion
    def dotenv(fn=".env", verbose=False):
    # ...
    if verbose:
        print("Loading environment variables from file: '%s'" % fn)
```

dotenv/Tiltfile
rrmistry: Same as the previous comment, I can think of adding a `verbose` or `debug` parameter in the parent function that would provide this diagnostic information.

What are your thoughts?

```suggestion
    def dotenv(fn=".env", verbose=False):
    # ...
    for lineNumber in range(len(lines)):
        # ...
        if verbose:
            print("Loading environment variable '%s'" % v[0])
        # ...
```

dotenv/Tiltfile
rrmistry: Thanks @nicks , good callout on the quotes

This is now fixed in https://github.com/tilt-dev/tilt-extensions/pull/498/commits/b64bec7bc92b7d6e12d2dd317b9617a784660973 and ready for review

atta:
Hi a nice thing to add would be: to also remove the \r on the end of each line if it exist for better windows support

rrmistry:
> Hi a nice thing to add would be: to also remove the \r on the end of each line if it exist for better windows support

The current PR relies on the StarLark function [`strip`](https://bazel.build/rules/lib/core/string#strip).

Looking at the documentation, it seems that when called without any parameters then it removes all whitespace. But it is not explicitly clear if it is `\n` or `\r` or both.

A simple task could be to add an explicit test case for this to make sure future breaking changes are caught (if any) for `\r` character support.

I can update this PR as soon as I get a chance. Thanks for pointing this out @atta !

PR #502, opened 2023-06-02
feat: k8s_yaml_glob extension for loading manifests by glob patterns.

gfranxman:
A simple extension to facilitate loading subsets of manifests by directory and filename patterns.

## Usage

First load the extension like:

    load('ext://k8s_yaml_glob', 'k8s_yaml_glob')
    
Then use it like this:

    # all of the .yaml files in the k8s directory.
    k8s_yaml_glob('./k8s/*.yaml')  
    
or like this:

    # all of the dev-*.yaml files in the k8s directory and subdirectories.
    k8s_yaml_glob('./k8s/**/dev-*.yaml')

### DIFF
--- a/k8s_yaml_glob/Tiltfile
+++ b/k8s_yaml_glob/Tiltfile
@@ -1,0 +1,16 @@
+# -*- mode: Python -*-
+
+def k8s_yaml_glob(*glob_patterns, **kwargs):
+    """
+    Create/apply list of Kubernetes YAML files matching the given glob patterns.
+    """
+
+    is_windows = True if os.name == "nt" else False
+    python_cmd = 'py -3' if is_windows else 'python3'
+
+    for glob_pattern in glob_patterns:
+        script = 'import glob; print("|".join( p for p in glob.glob(\"' + glob_pattern + '")))'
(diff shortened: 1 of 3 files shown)


k8s_yaml_glob/Tiltfile
```diff
@@ -0,0 +1,16 @@
...
+def k8s_yaml_glob(*glob_patterns, **kwargs):
+    """
+    Create/apply list of Kubernetes YAML files matching the given glob patterns.
+    """
+
+    is_windows = True if os.name == "nt" else False
+    python_cmd = 'py -3' if is_windows else 'python3'
+
+    for glob_pattern in glob_patterns:
+        script = 'import glob; print("|".join( p for p in glob.glob(\"' + glob_pattern + '")))'
```
nicks, line 12: i'm a little worried about the escaping here, but i guess we can add that when somebody needs it...

nicks (approved):
lgtm!

k8s_yaml_glob/Tiltfile
gfranxman: You're right, that vestigial escape is unnecessary.

PR #504, opened 2023-06-07
Validate flag for helm_remote extension

mhmxs:
I have found an issue with one of our Helm charts. 

`Error: chart requires kubeVersion: >=1.21.0-0 which is incompatible with Kubernetes v1.20.0\n\nUse --debug flag to render out invalid YAML`

My Kubernetes version is `1.25.3`, so I suspected `helm`. Here is the solution: https://github.com/helm/helm/issues/10086#issuecomment-909595139

But the case is Tilt built in `helm` function doesn't support validate flag, so this change uses a custom `template` command which optionally validate-able.

### DIFF
--- a/helm_remote/Tiltfile
+++ b/helm_remote/Tiltfile
@@ -54,7 +54,10 @@ metadata:
 #   =====================================
 
 
-def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], set=[], namespace='', version='', username='', password='', allow_duplicates=False, create_namespace=False):
+def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], set=[], namespace='', version='', username='', password='', allow_duplicates=False, create_namespace=False, validate=False):
+    values = [values] if type(values) == type('') else values
+    set = [set] if type(set) == type('') else set
+
     # ======== Helper methods
     def get_local_repo(repo_name, repo_url):
         # if no repos are present, helm exit code is >0 and stderr output buffered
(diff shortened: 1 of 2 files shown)


helm_remote/Tiltfile
```diff
@@ -178,7 +181,21 @@ def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], se
...
+
+    if len(set) != 0:
+        for s in set:
+            helm_tpl_cmd.extend(['--set', s])
+
+    if validate:
+        helm_tpl_cmd.append('--validate')
+
+
+    yaml = local(build_helm_command(helm_tpl_cmd, None, None))
```
nicks, line 198: i'm not sure that `build_helm_command` really adds much if you're not using the other args :joy:

nicks (approved):
lgtm!

nicks:
This had to be rolled back because it broke namespace injection. https://github.com/tilt-dev/tilt-extensions/issues/506. 

Sorry, I should have caught this. The `helm()` built-in does its own namespace injection because `helm template` does not. There's a longer history on this here: https://github.com/helm/helm/issues/5465

i think the best way to move this forward is to add the `validate` argument to the built-in `helm()` function. would you be willing to put together a pr for that?

mhmxs:
@nicks Thanks to catch the problem :) Does it make sense to reopen the PR with an appended  namespace?

yaml = local(`namespace` + build_helm_command(helm_tpl_cmd, None, None))

PR #505, opened 2023-06-07
Support --cache-to in kubectl_build

chrisjohnson:
https://github.com/vmware-tanzu/buildkit-cli-for-kubectl supports --cache-to, so we want to experiment and see how it behaves for us

I'm brand new to submitting requests to any tilt repos, so by all means point me toward any docs or guidelines I should be following if I'm doing this wrong please :)

### DIFF
--- a/kubectl_build/Tiltfile
+++ b/kubectl_build/Tiltfile
@@ -40,8 +40,6 @@ def kubectl_build(ref, context, build_args={}, dockerfile=None,
             kwargs["extra_tag"] = extra_tag
         if cache_from != None:
             kwargs["cache_from"] = cache_from
-        if cache_to != None:
-            kwargs["cache_to"] = cache_to
         if pull != None:
             kwargs["pull"] = pull
         if platform != None:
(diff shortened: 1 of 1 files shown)


kubectl_build/Tiltfile
```diff
@@ -40,6 +40,8 @@ def kubectl_build(ref, context, build_args={}, dockerfile=None,
             kwargs["extra_tag"] = extra_tag
         if cache_from != None:
             kwargs["cache_from"] = cache_from
+        if cache_to != None:
```
nicks, line 43: i don't think this will emit an error? this codepath is building up the args to docker_build, and docker_build does not support cache_to

chrisjohnson: I misunderstood the code but looking closer I agree. This seems to be the fallback logic for making kubectl_build pass through to docker_build if your specific build configuration isn't compatible with kubectl_build. I'll remove this change since my concern is the other direction (I want to add cache_to only for kubectl_build).

Are the CI tests attached here enough to feel confident that my change doesn't break typical use cases? What other kind of testing evidence may be needed for this PR to be merged?

nicks (approved):
lgtm!

kubectl_build/Tiltfile
nicks: ya, this is fine. there are tests but they're more for "make sure the basic use-cases work end to end" rather than for "test every combination of flags"

PR #511, opened 2023-06-24
Fix kubectl_build command

sambonbonne:
I had an issue with the `kubectl_build` extension: `kubectl build <context> …` made the CLI considere the `<context>` is the subcommand (as `kubectl build` has multiple commands: `create`, `rm`, `ls`…).

I fixed it by changing the `kubectl build …` command to `kubectl build build …`.


sambonbonne:
I'm sorry, it seems I had a wrong configuration using `kubectl build` as an alias to `kubectl buildkit`. I will fix it on my side.

Sorry for the useless PR.

PR #515, opened 2023-07-11
feat(procfile): allow passing kwargs to local_resource

Djiit:
Hi team,

Small change to allow passing kwargs to local_resource. Enable things like Labels or AutoInit=False on Procfile resources.

I'm not sure how to add tests, if needed. Happy to have some guidances.

Cheers

### DIFF
--- a/procfile/Tiltfile
+++ b/procfile/Tiltfile
@@ -1,4 +1,4 @@
-def procfile_resources(procfile='Procfile', set_port=False, starting_port=5000):
+def procfile_resources(procfile='Procfile', set_port=False, starting_port=5000, **kwargs):
   """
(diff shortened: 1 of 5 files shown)


nicks:
ya, can you add a test? if you put a test/test.sh script in the extension, the ci will pick it up. here's an example - https://github.com/tilt-dev/tilt-extensions/blob/master/local_output/test/test.sh

nicks:
hmmm your test failed when i tried them. i believe the problem is that procman / tilt expects your procfile to run servers that don't exit, and treats exit as failure. try something like `echo foo && sleep 100`

Djiit:
Alright, looks better now, thanks for the help :)

PR #524, opened 2023-08-31
adding `pod_readiness` flag to `helm_resource`

CSberger:
Fixes an issue where helm_resources pointing to helm charts without pods would never be marked as ready.


nicks (approved):
lgtm!

PR #525, opened 2023-09-01
helm_remote: fix pulling oci helm charts on windows

saschahemleb:
previously it would create a target path like
"..\\AppData\\Local\\tilt-dev\\.helm\\oci:\\repo.net:1234\\chart\\latest" which is not a valid path on windows. so now we just strip every :


nicks (approved):
lgtm!

PR #529, opened 2023-10-03
Create a showWarningsIfMissing flag to allow suppressing dotenv warnings

psigen:
This allows the ability to suppress a warning if a dotenv file is missing.

### DIFF
--- a/dotenv/Tiltfile
+++ b/dotenv/Tiltfile
@@ -1,8 +1,9 @@
-def dotenv(fn=".env", verbose=False, showValues=False):
+def dotenv(fn=".env", verbose=False, showValues=False, showWarningIfMissing=True):
 
     # Check if file exists
     if not os.path.exists(fn):
-        warn("Environment variables file not found: '%s'" % fn)
+        if showWarningIfMissing:
+            warn("Environment variables file not found: '%s'" % fn)
         return
 
     if verbose:


dotenv/Tiltfile
```diff
@@ -1,8 +1,9 @@
-def dotenv(fn=".env", verbose=False, showValues=False):
+def dotenv(fn=".env", verbose=False, showValues=False, showWarningIfMissing=True):
 
     # Check if file exists
     if not os.path.exists(fn):
-        warn("Environment variables file not found: '%s'" % fn)
+        if showWarningIfMissing:
+            warn("Environment variables file not found: '%s'" % fn)
```
psigen, line 6: Should I add a `watch_file()` here to re-trigger a build if the file is created?
The documentation is not clear about how `watch_file()` responds to missing files that are created after Tiltfile evaluation.
https://docs.tilt.dev/api.html#api.watch_file

dotenv/Tiltfile
```diff
@@ -1,8 +1,9 @@
-def dotenv(fn=".env", verbose=False, showValues=False):
+def dotenv(fn=".env", verbose=False, showValues=False, showWarningIfMissing=True):
```
psigen, line 1: Why are all the variables in this extension `camelCase`?  Isn't the Starlark norm for parameters `snake_case`?

Anyway, I left it as-is and copied the convention for my new flag.

nicks:
i feel like this makes the api unnecessarily complex. if you don't want the warning, you can do:

```
if os.path.exists('.env'):
  dotenv()
```

psigen:
That's what I'm doing right now.  I have a question though: do we also need to `watch_file()` to allow live-reload if the file shows up while Tilt is running?

PR #530, opened 2023-10-07
Support running helm dependency update

MattHodge:
Adds a new parameter to the `helm_resource` extension to allow triggering a `helm dependency update` on the given chart.

This is if you have a chart with file based dependencies eg:

```yaml
apiVersion: v2
name: foo
type: application
version: 0.0.1

dependencies:
  - name: "bar"
    repository: file://../../templates/bar
    version: 0.0.1
```

Without this, when you make changes to the dependent chart `bar`, they will not be shown in the rendered helm chart.

I tried just passing `--dependency-update` flag to the `helm install` command, but this doesn't work.


nicks (approved):
LGTM!

PR #538, opened 2023-12-07
add base64 and dict_merge extension

gebinic:
Added an extension which encodes and decodes a string. Supported unix and windows systems.

EDIT: add dict_merge extension

### DIFF
--- a/base64/Tiltfile
+++ b/base64/Tiltfile
@@ -1,0 +1,9 @@
+def encode_base64(content):
+    command_value = 'echo {} | base64'.format(content)
+    command_bat_value = ['powershell', '[convert]::ToBase64String([Text.Encoding]::UTF8.GetBytes(\'{}\'))'.format(content)]
+    return str(local(command=command_value, quiet=True, command_bat=command_bat_value, echo_off=True)).strip()
+
+def decode_base64(content):
+    command_value = 'echo {} | base64 -d'.format(content)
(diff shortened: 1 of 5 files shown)


nicks:
thanks for the contribution!

i'm fine with the `base64` extension

i'm nervous about having the `dict_merge` extension in the main extensions repo, because:
- starlark already has a dictionary union operator (see: https://github.com/bazelbuild/starlark/blob/master/spec.md#dictionaries)
- historically, deep merge has been pretty controversial, because it's hard to make the behavior intuitive

gebinic:
We distinguish between `deploy_config.yml` ("global" configuration file) and `local_config.yml` (git ignored file, only local user configuration file). Due to our configuration model, it is extremely difficult to merge these configuration files manually, among other things with the dictionary union operator. Therefore, we decided to introduce the `dict_merge` function, which simply merges all configurations deeply. I think this could be a use case that also affects other tilt users.

base64/Tiltfile
```diff
@@ -0,0 +1,9 @@
+def encode_base64(content):
+    command_value = 'echo {} | base64'.format(content)
```
remram44, line 2: With longer strings, base64 will insert newlines every 76 characters. `base64 -w 0` avoids that.

nicks:
i think we'd accept a pr for just the `base64` but wouldn't accept `dict_merge`, for the reasons stated above. closing this pr, but feel free to repopen for the `base64` part

Issue #540, opened 2023-12-22
`restart_process` fails with permission denied


thejan2009:
`docker_build_with_restart` has been failing for me recently, but I've found out what's the issue - even though `/tmp/.restart-proc` has permissions `666`, the `/tmp` folder has [sticky bit](https://en.wikipedia.org/wiki/Sticky_bit) enabled and the containers run with `root` user instead of `1001` that's the owner of `/tmp/.restart-proc`.

```
Will copy 1 file(s) to container: [api-fb9658fb7-pg8pd/api]
- '~/code/go/build/api' --> '/bin/api'
[CMD 1/1] sh -c date > /tmp/.restart-proc
sh: can't create /tmp/.restart-proc: Permission denied
  → Failed to update container api-fb9658fb7-pg8pd/api: executing on container 818f0243f0: command "date > /tmp/.restart-proc" failed with exit code: 1
Build Failed: executing on container 818f0243f0: command "date > /tmp/.restart-proc" failed with exit code: 1
```
I've found two temporary workarounds, but they won't scale:
 - `kubectl exec` into the pod and toggle `/tmp` sticky bit
 - `kubectl exec` into the pod and `chown /tmp/.restart-proc`
 
 Any other ideas on how to solve this? Could the `.restart-proc` file be moved to `/tmp/.restart/proc` and sticky bit toggled on `tmp/.restart` folder?

nicks:
The fix was rolled back due to the regression reported in https://github.com/tilt-dev/tilt-extensions/issues/544, sorry

thejan2009:
Hey, I took a look at the issue there, not sure how to proceed.

```
       → /bin/sh: 1: [mkdir,: not found
...
process "/bin/sh -c [\"mkdir\", \"-p\", \"\\tmp\\.restart\"]"
```

It seems for some reason docker builder garbles the parameters for this RUN command in exec form. And that path is also not correct. We can work around missing `mkdir` by using `WORKDIR` dockerfile instruction. 

But the `os.path.dirname` output will still be a windows-style path that will, in the best case, create a useless directory. I think here we'll always need a linux-style `dirname` function. Any idea how to get there in Tiltfile? In golang, this would mean using `path.Dir` instead of [`filepath.Dir`](https://github.com/tilt-dev/tilt/blob/f7942718e2ed699841d21ed1665272c410d81540/internal/tiltfile/os/os.go#L189).

nicks:
[...]

PR #541, opened 2023-12-22
fix(restart_process): work around sticky /tmp

thejan2009:
Sometimes the container root filesystem has sticky bit enabled and restart commands are ran with a different user. We can work around that by creating an intermediate folder that has "0777" permissions.

Refs #540


nicks (approved):
lgtm! thanks!

Issue #544, opened 2024-01-03
Building Docker Image Failing with 'restart_process' extension on Windows


djhelmick:
Tilt: v0.33.10, built 2023-12-15
System: windows-amd64

Beginning this afternoon, I've started seeing the below error when attempting to run my Tiltfile using the `restart_process` extension.

```
STEP 3/5 — Building Dockerfile: [***]
Building Dockerfile for platform linux/amd64:
  
  FROM tiltdev/restart-helper:2021-11-03 as restart-helper
  
  FROM localhost:63333/[***-tilt_docker_build_with_restart_base:tilt-612fc2746907dfe5
  RUN ["mkdir", "-p", "\tmp\.restart"]
  RUN ["touch", "/tmp/.restart/proc"]
  RUN ["chmod", "666", "/tmp/.restart/proc"]
  COPY --from=restart-helper /tilt-restart-wrapper /
  COPY --from=restart-helper /entr /
  ENTRYPOINT ["/tilt-restart-wrapper", "--watch_file=/tmp/.restart/proc", "dotnet", "[***.dll"]


     Building image
     [restart-helper 1/1] FROM docker.io/tiltdev/restart-helper:2021-11-03@sha256:74d89eac884107e73e2f947a72e702750c92b16e2fae3b1da1ac36c14e79c5ff
     [stage-1 1/6] FROM localhost:63333/[***-tilt_docker_build_with_restart_base:tilt-612fc2746907dfe5
     [stage-1 2/6] RUN ["mkdir", "-p", "\tmp\.restart"]
       → /bin/sh: 1: [mkdir,: not found
     
     ERROR IN: [stage-1 2/6] RUN ["mkdir", "-p", "\tmp\.restart"]

Build Failed: ImageBuild: process "/bin/sh -c [\"mkdir\", \"-p\", \"\\tmp\\.restart\"]" did not complete successfully: exit code: 127
```

I hadn't made any changes to the Tiltfile so at first I was at a bit of a loss as to why it suddenly started yelling about `mkdir` during this step. After looking through the repos, however, I noticed that the PR https://github.com/tilt-dev/tilt-extensions/pull/541 was accepted about an hour ago which adds in that `RUN` step.

![image](https://github.com/tilt-dev/tilt-extensions/assets/27941531/4231f456-f554-4e4f-8644-7fd33262253b)

[...]

PR #545, opened 2024-01-03
Revert "fix(restart_process): work around sticky /tmp"

nicks:
Reverts tilt-dev/tilt-extensions#541


nicks:
Rolling back due to reports of breakage in https://github.com/tilt-dev/tilt-extensions/issues/544

PR #547, opened 2024-01-04
fix: pass through 'platform' arg to nerdctl

lostrouter:
The `nerdctl` extension was not passing through the `platform` argument to the under lying command. 

tested locally with the following example

```
nerdctl_build(
  ref='my-docker-image', 
  context='.',
  platform='amd64'
)
```

output observed in the Tilt UI

```
Running cmd: nerdctl --namespace k8s.io build --platform amd64 -t "${EXPECTED_REF}" .
```


nicks (approved):
lgtm!

PR #548, opened 2024-01-05
fix(restart_process): work around sticky /tmp

thejan2009:
Another attempt to fix this. Changes from previous iteration:
- use string.rfind('/') to extract folder instead of os.path.dirname
- join RUN commands

Disclaimer, I've not been able to test this on Windows.

If `mkdir` should not be used, this is a WORKDIR solution:

``` diff
     FROM tiltdev/restart-helper:2021-11-03 as restart-helper
 
-    FROM {ref}
+    FROM {ref} as workdir
+    WORKDIR "{folder}"
     RUN ["touch", "{file}"]
     RUN ["chmod", "666", "{file}"]
+
+    FROM {ref}
+    COPY --from=workdir "{file}" "{file}"
     COPY --from=restart-helper /tilt-restart-wrapper /
     COPY --from=restart-helper /entr /
```

Sometimes the container root filesystem has sticky bit enabled and restart commands are ran with a different user. We can work around that by creating an intermediate folder that has "0777" permissions.

The folder name is parsed with `string.rfind` because os.path.dirname won't work on Windows.

Refs #540

### DIFF
--- a/restart_process/Tiltfile
+++ b/restart_process/Tiltfile
@@ -1,4 +1,4 @@
-RESTART_FILE = '/tmp/.restart-proc'
+RESTART_FILE = '/tmp/.restart/proc'
 TYPE_RESTART_CONTAINER_STEP = 'live_update_restart_container_step'
 
 KWARGS_BLACKLIST = [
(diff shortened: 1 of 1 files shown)


nicks (approved):
lgtm!

nicks:
unfortunately, we got a report that this broke things again, due to lack of `sh` in the base container. reverting...

nicks:
https://kubernetes.slack.com/archives/CESBL84MV/p1704483542084299

PR #552, opened 2024-01-16
Add pypiserver extension

korbajan:


korbajan:
extension is using `docker_compose` (https://github.com/korbajan/tilt-extensions/blob/pypiserver/pypiserver/Tiltfile#L41) but tests are failing with: `error: exec: "docker-compose": executable file not found in $PATH` (https://app.circleci.com/pipelines/github/tilt-dev/tilt-extensions/1524/workflows/e68dbb44-985c-4af2-93e1-822888b71262/jobs/3638)

nicks:
ya, the ci doesn't have `compose` installed, working on it - https://github.com/tilt-dev/tilt-extensions/pull/553

pypiserver/Tiltfile
```diff
@@ -0,0 +1,66 @@
+EXT_PATH = os.getcwd()
+PYPI_PORT_VAR_NAME = 'PYPISERVER_PORT' # docker compose purposes
+DEFAULT_PYPI_PORT_VALUE = '8222'
+DEFAULT_CACHE_DIR = '.tilt-cache'
```
nicks, line 4: i'm not sure it makes much sense to call this a `tilt cache`. or use `TILT_CACHE_DIR` to configure it

pypiserver/Tiltfile
```diff
@@ -0,0 +1,66 @@
...
+DEFAULT_PYPI_CONTAINER_NAME = 'tilt-pypiserver'
+PYPI_CONTAINER_VAR_NAME = 'PYPI_CONTAINER_NAME' # docker compose purpose
+PYPISERVER_DATA_CACHE_DIR_VAR_NAME = 'PYPISERVER_PACKAGES_DATA_CACHE' # docker compose purposes
+PYPISERVER_DATA_CACHE_DIR = 'pypiserver'
+
+PYTHON_CMD = 'python3'
+
+def _root_tiltfile_dir():
```
nicks, line 12: this is slightly incorrect, can you copy the implementation from `git_resource`?

nicks (commented):
can you rebase your pr to get the new ci image?

pypiserver/Tiltfile
korbajan: But the one from `git_resource` does not work ([test](https://app.circleci.com/pipelines/github/tilt-dev/tilt-extensions/1529/workflows/d4b7b1bc-5f26-49d8-8d00-1defdfb37023/jobs/3651)), it passes the tests only because there is [this](https://github.com/korbajan/tilt-extensions/blob/pypiserver/git_resource/test/Tiltfile#L1) setting in the test/Tiltfile :)

pypiserver/Tiltfile
korbajan: so `.tilt-cache` is just a fallback, default value when there is no TILT_CACHE_DIR

nicks:
i'll merge this and fix up the load logic to match the other ones async

PR #554, opened 2024-01-26
Add base64 extension

remram44:
It looks like the PowerShell version strips newlines inside the content, I am not sure it can be avoided. Unless we want to build the function natively, which I think we should.

Closes #538


nicks (approved):
Lgtm!

PR #556, opened 2024-02-02
Podman: add restart function

ianb-mp:
Add `podman_build_with_restart()` function to Podman extension.

@nicks fyi

### DIFF
--- a/podman/README.md
+++ b/podman/README.md
@@ -17,11 +17,36 @@ Use Podman (https://podman.io/) to build images for Tilt.
 - **live_update**: Set of steps for updating a running container
       (see https://docs.tilt.dev/live_update_reference.html)
 
+### `podman_build_with_restart(ref: str, context: str, entrypoint: str, extra_flags?: [str])`
+
+- **ref**: The name of the image to build. Must match the image
+   name in the Kubernetes resources you're deploying.
+- **context**: The build context of the binary to build. Expressed as a file path.
+- **entrypoint**: The command to be (re-)executed when the container starts or when a live_update is run.
+- **ignore**: Changes to the given files or directories do not trigger rebuilds.
+      Does not affect the build context.
+- **extra_flags**: Extra flags to pass to podman build.
+- **live_update**: Set of steps for updating a running container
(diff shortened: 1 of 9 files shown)


nicks (approved):

arlyon:
This seems to invoke docker still, no?

https://github.com/tilt-dev/tilt-extensions/blob/master/restart_process/Tiltfile

Should I open an issue?

ianb-mp:
@arlyon I'm not sure what you mean by 'invoke docker'. Can you provide more detail? This PR passes a `command` parameter [here](https://github.com/ianb-mp/tilt-extensions/blob/52cdda16b4fc0b6702046718a856ccd5eaae05ab/podman/Tiltfile#L73-L75) which explicitly uses podman.

arlyon:
Sure. `podman_build_with_restart` will unconditionally call into docker to build the image that hosts the `tilt-restart-wrapper`. The call graph is as follows:
- podman_build_with_restart https://github.com/tilt-dev/tilt-extensions/blob/master/podman/Tiltfile#L44C5-L44C30
- custom_build_with_restart https://github.com/tilt-dev/tilt-extensions/blob/master/restart_process/Tiltfile#L110
- _helper https://github.com/tilt-dev/tilt-extensions/blob/master/restart_process/Tiltfile#L27C5-L27C12
- docker_build

ianb-mp:
Ah I see what you mean. TBH I don't understand the internals of tilt and the multiple stages. I do know that `podman_build_with_restart` is working for me in an environment where everything outside the k8s cluster is Podman only. Is the later `docker_build` happening inside a container? And if so, would it matter if it's docker and not Podman?

nicks:
@arlyon Tilt communicates with docker over the unix socket / REST API. Podman claims to offer REST API compatibility, so should work fine if you have the environment set up correctly. if it's not working for you, please open an issue with repro steps.

arlyon:
Apologies for the late reply. I don't have podman running in daemon mode (it is daemonless by default) but this can be an ok workaround. I will open an issue with a repro. Thanks for all the amazing work you guys are doing! 💞

PR #562, opened 2024-03-15
[kubebuilder] Remove GOARCH to make it work on the Apple Silicon（M1/M2) env

LittleWat:
## change
- [kubebuilder] Remove GOARCH to make it work on the Apple Silicon（M1/M2）env

## test
- I confirmed that this works on both the linux amd64 env and the darwin arm64 env.

### DIFF
--- a/kubebuilder/Tiltfile
+++ b/kubebuilder/Tiltfile
@@ -36,7 +36,7 @@ def kubebuilder(DOMAIN, GROUP, VERSION, KIND, IMG='controller:latest', CONTROLLE
     
     # build to tilt_bin beause kubebuilder has a dockerignore for bin/
     def binary():
-        return 'CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o tilt_bin/manager cmd/main.go'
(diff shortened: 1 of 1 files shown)


kubebuilder/Tiltfile
```diff
@@ -36,7 +36,7 @@ def kubebuilder(DOMAIN, GROUP, VERSION, KIND, IMG='controller:latest', CONTROLLE
     
     # build to tilt_bin beause kubebuilder has a dockerignore for bin/
     def binary():
-        return 'CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o tilt_bin/manager cmd/main.go'
+        return 'CGO_ENABLED=0 GOOS=linux GO111MODULE=on go build -o tilt_bin/manager cmd/main.go'
```
nicks, line 39: hmmm...i'm not sure this is quite right. you really want this to match the arch of the cluster, not the local machine.

you might be able to get the right arch with something like

```
$(tilt get cluster default -o=jsonpath --template="{.status.arch}")
```

PR #566, opened 2024-04-04
secrets: add helpers for docker registry and ECR

nicks:
Signed-off-by: Nick Santos <nick.santos@docker.com>


nicks:
i found this useful in a few other tiltfiles i was writing...probably also want to add some helpers to attach these secrets to serviceaccounts

nicksieger (approved):
Nice!

Issue #528, opened 2023-09-25
`helm_resource` KeyError


StephanMeijer:
```
     Running cmd: python3 /home/steve/.local/share/tilt-dev/tilt_modules/github.com/tilt-dev/tilt-extensions/helm_resource/helm-apply-helper.py
     Traceback (most recent call last):
       File "/home/steve/.local/share/tilt-dev/tilt_modules/github.com/tilt-dev/tilt-extensions/helm_resource/helm-apply-helper.py", line 24, in <module>
         image = os.environ['TILT_IMAGE_%s' % i]
                 ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
       File "<frozen os>", line 679, in __getitem__
     KeyError: 'TILT_IMAGE_3'
```

Probably caused due to using multiple same Docker refs:

```py
load('ext://helm_resource', 'helm_resource', 'helm_repo')

docker_build('dev-dcs-php-generic',
             '.',
             pull=True,
             dockerfile='.ci/docker/php-generic/Dockerfile')

docker_build('dev-dcs-php-pandoc',
             '.',
             dockerfile='.ci/docker/php-libreoffice/Dockerfile',
             build_args={'BASE_IMAGE': 'dev-dcs-php-generic'})

docker_build('dev-dcs-php-libreoffice',
             '.',
             dockerfile='.ci/docker/php-libreoffice/Dockerfile',
             build_args={'BASE_IMAGE': 'dev-dcs-php-generic'})

helm_repo('nldoc', 'https://gitlab.com/api/v4/projects/50351992/packages/helm/stable')
helm_resource(
    name='dev-dcs',
    chart='nldoc/nldoc-stack',
    image_deps=[
        'dev-dcs-php-generic',
        'dev-dcs-php-generic',
        'dev-dcs-php-generic',
        'dev-dcs-php-pandoc',
        'dev-dcs-php-libreoffice',
    ],
    image_keys=[
        ("nldoc-conversion.php-dcs-inbound.image.repository",
         "nldoc-conversion.php-dcs-inbound.image.tag"),
        ("nldoc-conversion.php-dcs-publicatietool-inbound.image.repository",
         "nldoc-conversion.php-dcs-publicatietool-inbound.image.tag"),
        ("nldoc-conversion.php-dcs-outbound.image.repository=dev-dcs-php-generic",
         "nldoc-conversion.php-dcs-outbound.image.tag"),
[...]

PR #569, opened 2024-04-10
helm_resource: allow images to be injected twice

nicks:
fixes https://github.com/tilt-dev/tilt-extensions/issues/528

Signed-off-by: Nick Santos <nick.santos@docker.com>

### DIFF
--- a/helm_resource/Tiltfile
+++ b/helm_resource/Tiltfile
@@ -95,20 +95,15 @@ def helm_resource(
   if flags:
     apply_cmd.extend(flags)
 
+  dupe_deps = {}
   for i in range(len(image_deps)):
+    if image_deps[i] in dupe_deps:
+      fail(('helm_resource: image_deps should be unique. Found duplicate: %s\n' % image_deps[i]) +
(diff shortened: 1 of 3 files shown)


nicks:
a lot of what i struggled with here was what the right "interface" for users should be.

i _think_ we want to error out if a user adds duplicate elements to `image_deps`, but i could be convinced otherwise

helm_resource/Tiltfile
```diff
@@ -95,20 +95,16 @@ def helm_resource(
   if flags:
     apply_cmd.extend(flags)
 
+  dupe_deps = {}
   for i in range(len(image_deps)):
+    env['TILT_IMAGE_DEP_%s' % i] = image_deps[i]
```
nicksieger, line 100: Is this var used or supposed to be used? I also notice that the help apply helper has a reference to `TILT_IMAGE_%s`, are they supposed to be the same?

nicksieger (approved):
Approach looks like a good balance of complexity and flexibility. Just one question

helm_resource/Tiltfile
nicks: ah, good call, i think we can remove this

Issue #572, opened 2024-04-16
helm_remote enters infinite loop on Windows, with tilt_modules folder missing


Venryx:
The logic for traversing up and finding the Tiltfile folder is hard-coded for Linux -- well, the part that ensures an infinite loop doesn't occur.

The function in question: (from: https://github.com/tilt-dev/tilt-extensions/blob/027c37bb3dc0ce11ee62eadac8c37d3742b53f6f/helm_remote/Tiltfile#L19C1-L28C41)
```
def _find_root_tiltfile_dir():
    # Find top-level Tilt path
    current = os.path.abspath('./')
    while current != '/':              # <-- this line
        if os.path.exists(os.path.join(current, 'tilt_modules')):
            return current


        current = os.path.dirname(current)


    fail('Could not find root Tiltfile')
```

Notice this line: `while current != '/':`

On Windows, this condition never evaluates to false, even after reaching the drive root, causing an infinite loop to occur if the "tilt_modules" folder doesn't exist (which is normal in my repo for a fresh clone; it is populated only after Tilt has had a chance to fully execute).

Note: Noticed this issue when a freelancer found tilt inexplicibly failing to run to the end of the script file; we identified this as the cause by using print-lines/bisection.

nicks:
Hmmmm...i wasn't able to reproduce this problem.

Can you share what version of Tilt you're running?

Here's what I tried

1) Started a Windows box and opened a Cmd terminal
2) Opened the helm_remote tests
3) Updated the tiltfile to:

```
load('ext://helm_remote', 'helm_remote')

helm_remote('memcached', repo_url='https://charts.bitnami.com/bitnami')

docker_build('helm-remote-test-verify', '.')
k8s_yaml('job.yaml')
k8s_resource('helm-remote-test-verify', resource_deps=['memcached'])
```

And everything worked!


re: "the "tilt_modules" folder doesn't exist (which is normal in my repo for a fresh clone" -- that doesn't really make sense to me -- you can't even run this code unless you've downloaded it, and to downloading it puts it in a tilt_modules dir.

Venryx:
> re: "the "tilt_modules" folder doesn't exist (which is normal in my repo for a fresh clone" -- that doesn't really make sense to me -- you can't even run this code unless you've downloaded it, and to downloading it puts it in a tilt_modules dir.

[...]

PR #573, opened 2024-04-19
helm_remote: update directory crawling to be a bit more idiomatic

nicks:
pointed out in
https://github.com/tilt-dev/tilt-extensions/issues/572

to be clear, i don't think this will actually
fix the bug reported there. it's likely a deeper
issue. but i agree that we shouldn't be hard-coding
'/' to terminate the loop


nicksieger (approved):

nicksieger:
Worth a try, seems like it might help.

PR #581, opened 2024-05-08
logger: add logger extension

gebinic:
For larger tilt projects with a lot of `print` function usages, it might be useful to have a more readable and colorful text output in the tilt ui. This approach makes things like debugging much easier. This logger is inspired by the [Maven](https://maven.apache.org/) logger.


nicks (approved):
lgtm!

PR #583, opened 2024-05-10
Forward signals in restart wrapper

### LINKED ISSUE #577: custom_build_with_restart does not pass signals to wrapped process
I'm using `custom_build_with_restart` in conjunction with a k8s_resource. I've been testing graceful shutdowns, and noticed that the custom_build that is created does not actually forward signals to the wrapped process. Docker relies on the entrypoint having pid=1, but with the wrapper, the process that actually needs to handle the signals no longer has that pid.

jscheel:
This PR fixes #577 by forwarding signals from the tilt-restart-wrapper to the entrypoint that is wrapped.


jscheel:
This test is supposed to have a failing job, so not sure what needs to be done to update the circleci test run.

jscheel:
Converted this to draft until I figure out the specs

jscheel:
Ok, so this spec fails on master as well. It appears that the live_update does to run on `tilt ci`. Anyone know anything about that?

restart_process/tilt-restart-wrapper.go
```diff
@@ -52,6 +53,18 @@ func main() {
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 
+	// Set up a signal forwarding handler
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs)
+	go func() {
+		for {
+			sig := <-sigs
+			if cmd.Process != nil {
```
nicks, line 62: i don't think checking cmd.Process like this is thread-safe?

we probably need something like:

```
err := cmd.Start()

// set up process control

err = cmd.Wait()
```

tmc:
Just on ping here, as this would be helpful.

jscheel:
Sorry, I’m traveling in Vietnam right now. I can fix up in a couple of min
as soon as I get back.

tmc:
I pushed a set of improvements here: https://github.com/tmc/tilt-extensions/tree/forward-signals-in-restart-wrapper

PR #584, opened 2024-05-17
base64: fix error on macOS

gebinic:
- the base64 binary on macOS has no "-w" cli option. there is an alternative "-b" which has "0" as default, so we pass "-w 0" only if we are NOT on macOS
- implement local_output module

### DIFF
--- a/base64/Tiltfile
+++ b/base64/Tiltfile
@@ -1,17 +1,29 @@
+load('ext://local_output', 'local_output')
+
 def encode_base64(content):
-    return str(local(
-        command=['base64', '-w', '0'],
+    base64_cmd = ['base64']
+    # the base64 binary on macOS has no "-w" cli option. there is an alternative "-b" which has "0" as default, so we pass "-w 0" only if we are NOT on macOS
+    # see https://stackoverflow.com/a/46464081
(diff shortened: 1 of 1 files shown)


base64/Tiltfile
```diff
@@ -1,17 +1,29 @@
+load('ext://local_output', 'local_output')
+
 def encode_base64(content):
-    return str(local(
-        command=['base64', '-w', '0'],
+    base64_cmd = ['base64']
+    # the base64 binary on macOS has no "-w" cli option. there is an alternative "-b" which has "0" as default, so we pass "-w 0" only if we are NOT on macOS
```
nicks, line 5: i don't believe this is accurate -- the problem you're running into is a difference between BSD base64 and GNU base64
https://man.freebsd.org/cgi/man.cgi?query=base64&apropos=0&sektion=0&manpath=FreeBSD+5.3-RELEASE+and+Ports&format=html
https://www.gnu.org/software/coreutils/manual/html_node/base64-invocation.html#base64-invocation

if you do a `brew install coreutils`, you can install GNU base64 on macOS just fine

instead, maybe the way to do this is to check `base64 --version`?

gebinic: Many thanks for this hint! I replaced the os check with a bsd/gnu check in my last commit.

PR #591, opened 2024-07-16
helm_resource: `links` argument

noelleleigh:
This PR adds the `links` argument from [`k8s_resource()`](https://docs.tilt.dev/api#api.k8s_resource) to `helm_resource()` so that clickable URLs can be added to the resource.


nicks (approved):
lgtm!

PR #592, opened 2024-07-24
feat: Add file extension (read/write)

freehuntx:
This PR adds a file extension which allows read/write of files.

### DIFF
--- a/file/Tiltfile
+++ b/file/Tiltfile
@@ -1,0 +1,21 @@
+def read_file(path):
+  if os.name == "posix": # Linux/Mac implementation
+    return str(local("cat {}".format(path), echo_off=True, quiet=True))
+  elif os.name == "nt": # Windows implementation
+    path = path.replace("/", "\\")
+    return str(local("type {}".format(path), echo_off=True, quiet=True))
+  else:
+    fail("Unknown system architecture: " + os.name)
+
+def write_file(path, content):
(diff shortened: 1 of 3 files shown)


humb1t (approved):

file/Tiltfile
```diff
@@ -0,0 +1,21 @@
+def read_file(path):
```
nicks, line 1: hmmm...`read_file` is already a builtin, so i'm not sure what this adds - https://docs.tilt.dev/api.html#api.read_file

file/Tiltfile
```diff
@@ -0,0 +1,21 @@
...
+    path = path.replace("/", "\\")
+    return str(local("type {}".format(path), echo_off=True, quiet=True))
+  else:
+    fail("Unknown system architecture: " + os.name)
+
+def write_file(path, content):
+  content = str(content)
+
+  if os.name == "posix": # Linux/Mac implementation
+    local("echo '{}' > {}".format(content, path), echo_off=True, quiet=True)
```
nicks, line 14: this is going to cause escaping problems, so i'm not sure this is a good idea

freehuntx: Np i close it

file/Tiltfile
freehuntx: Its a bad MR i guess

PR #601, opened 2024-09-10
helm_resource: kubectl_cmd should define `--namespace` flag

SamWolfs:
If `namespace` is defined, the `--namespace` flag should be appended to the `kubectl_cmd` as well, otherwise the command will be unable to find the resources that are deployed to the non-default namespace.

### DIFF
--- a/helm_resource/helm-apply-helper.py
+++ b/helm_resource/helm-apply-helper.py
@@ -60,6 +60,7 @@ namespace = os.environ.get('NAMESPACE', '')
 if namespace:
   install_cmd.extend(['--namespace', namespace])
   get_cmd.extend(['--namespace', namespace])
+  kubectl_cmd.extend(['--namespace', namespace])
 
 install_cmd.extend([release_name, chart])
 get_cmd.extend([release_name])


nicks:
can you file a bug that gives:
1. specific repro steps
2. what you expect to happen
3. what actually happened
?

`helm_resource` already has a lot of tests for how namespaces should behave. 

The reason that the tests fail with your PR is because there are big differences `helm --namespace` flag does and what the `kubectl --namespace` flag does

SamWolfs:
Will do, I thought it was a quick fix that shouldn't impact anything else, but clearly I was wrong. I'll try to reproduce in the test.

SamWolfs:
@nicks I'll close the PR since it looks like I was a bit too fast, thinking I had a fix for a problem that didn't exist. Instead it was caused by CRLF line endings, which aren't supported by `namespacing.py`, and so explicitly setting the namespace on the `kubectl` command clearly fixed the local issue. If you think this is something that should be supported, I wouldn't mind creating an issue and reproduction for it.

nicks:
@SamWolfs oh interesting! ya, a repro case would help a lot! :pray:

PR #611, opened 2024-09-23
[helm_resource] Support specifying custom helm uninstall flags

mikesouza:
This is a simple PR that adds an optional argument `uninstall_flags` to the `helm_resource` extension to support the ability to pass custom flags to the `helm uninstall` command. This is similar to the existing `flags` argument which passes custom flags to the `helm upgrade --install` command.


nicks (approved):
lgtm!

PR #612, opened 2024-10-09
add helm_remote_yaml

sethpollack:
Adds `helm_remote_yaml`

Example use case would be generating the yaml with helm and then applying some transformations before installing.

### DIFF
--- a/helm_remote/Tiltfile
+++ b/helm_remote/Tiltfile
@@ -57,6 +57,27 @@ metadata:
 
 
 def helm_remote(chart, repo_url='', repo_name='', release_name='', values=[], set=[], namespace='', version='', username='', password='', allow_duplicates=False, create_namespace=False):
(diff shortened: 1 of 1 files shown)


nicks:
for other extensions, the API pattern we've used is to break this into two separate functions -- e.g., deployment_create and deployment_yaml - https://github.com/tilt-dev/tilt-extensions/tree/master/deployment

could we do the same thing here? so this would become helm_remote_yaml?

sethpollack:
Works for me. Just to be clear, add a new extension `helm_remote_yaml` and then update `helm_remote` to use `helm_remote_yaml` and call `k8s_yaml` with the output, right?

sethpollack:
@nicks Let me know if this is what you had in mind?

nicks (approved):
nice!

sethpollack:
Thanks!

PR #616, opened 2024-11-07
fix: adds option to disable helm_remote install_crds

### LINKED ISSUE #615: [helm_remote] install CRDs should be optional
https://github.com/tilt-dev/tilt-extensions/blob/5c7dd74880c2c1f6bc91a78628b1bf1eb0947a53/helm_remote/Tiltfile#L65

I have a helm chart value crds.enabled=false and was still seeing CRDs to be installed, it got me so confused with this strange "side effect". Can we make this line optional?

XBeg9:
fixes #615


nicks (approved):
LGTM

PR #617, opened 2024-11-21
Allow custom restart image for restart_process

jskrill:
For context, at my organization we do not have access to docker.io and need to use images mirrored in a custom registry (but not at the same image path). This PR allows one to set a custom image as the restart_image while preserving default behavior in a backward compatible manner.

### DIFF
--- a/restart_process/Tiltfile
+++ b/restart_process/Tiltfile
@@ -1,5 +1,6 @@
 RESTART_FILE = '/tmp/.restart-proc'
 TYPE_RESTART_CONTAINER_STEP = 'live_update_restart_container_step'
+DEFAULT_RESTART_IMAGE = "tiltdev/restart-helper:2024-06-06"
 
 KWARGS_BLACKLIST = [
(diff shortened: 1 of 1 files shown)


nicks:
sorry, i don't think we can accept this change.

This PR would make the extension unmaintainable -- any change to the extension image would break you, because we have no way of updating your forked image.

your best bet is to fork the extensions repo and pull the extension from your fork, here are docs on how to do this - https://docs.tilt.dev/extensions

jskrill:
Sorry, I'm not sure I follow this line. The extension is unusable today because the image cannot be pulled at all in an environment without access to docker.io and more generally, the public internet. While it is possible that some update to tilt-extensions could be incompatible with a custom image, it seems like a better option than having the tilt extension be non-functional in those environments.

nicks:
Yeah, I understand the problem.

I'm saying that the most common solution to this problem is to have a private extension repo that uses company-private images, which will be much more api-stable.

PR #620, opened 2024-11-29
Fix shellcheck CI job, fix/handle all shellcheck findings

dbast:
This fixes the shellcheck CircleCI job:
* The existing job does nothing as the CI images does not support DIND (docker in docker), see error  `docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?` in https://app.circleci.com/pipelines/github/tilt-dev/tilt-extensions/1678/workflows/c1a75955-880d-408c-8188-cfe7651a9085/jobs/4065
* That is why the shellcheck circle orb is used instead.
* Even if that would work, it would still have always run green as the code in the Makefile target does not propagate all the errors.
* The Makefile target is updated so that the errors propagate to the final exit code ... for local bug fixing, without running circle ci.
* The Make target is further updated to use the official upstream container `koalaman/shellcheck:stable`, which is not 7 years old.

This PR additional fixes / handles all the previous not shown shellcheck findings, to get the noise down and the PR green -> CI becomes actionable feedback for new PRs.


dbast:
yay. finally green after several iterations. Ready for review. thanks!

.circleci/config.yml
```diff
@@ -1,4 +1,6 @@
 version: 2.1
+orbs:
```
nicks, line 2: can you change it back to use `make shellcheck`?

we actually moved away from this orb -- it was important to us that the local Makefile and the circleci job ran the same code

https://github.com/tilt-dev/tilt-extensions/pull/223

you'll need to add a setup_remote_docker to make the docker commands work
https://circleci.com/docs/building-docker-images/

Makefile
```diff
@@ -7,5 +7,10 @@ publish-ci-image:
 	docker buildx build --push --pull --platform linux/amd64 -t docker/tilt-extensions-ci -f .circleci/Dockerfile .circleci
 
 shellcheck:
-	find . -type f -name '*.sh' -not -path "*/node_modules/*" -not -path "*/.git-sources/*" -not -path "*/.git/*" -exec \
-	    docker run --rm -it -e SHELLCHECK_OPTS="-e SC2001" -v $$(pwd):/mnt nlknguyen/alpine-shellcheck {} \;
+	LINT_ERRORS=0; \
+	IFS=$$'\n'; for shellscript in $$(git ls-files -z | xargs -0 file | grep "shell script" | cut -d: -f1 || echo ""); do \
```
nicks, line 11: rather than adding more complex shell scripting in the makefile, can you create a shell script in a `hack` directory? that's usually the kubernetes convention

Makefile
```diff
@@ -7,5 +7,10 @@ publish-ci-image:
...
+	   docker run --rm -v "$$PWD:/mnt" koalaman/shellcheck:stable "$${shellscript}" || LINT_ERRORS=$$((LINT_ERRORS+1)); \
```
nicks, line 12: does this `koalaman/shellcheck` support SHELLCHECK_OPTS?

get_obj/test/test.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-cd "$(dirname "$0")"
+cd "$(dirname "$0")" || exit
```
nicks, line 3: shouldn't this be `exit 1`?

syncback/krsync.sh
```diff
@@ -28,27 +28,31 @@ install_rsync_tilt() {
...
 
 find_rsync_path() {
     local rsync_path=--rsync-path=$RSYNC_TILT_PATH
     # see if we already have a compatible rsync on the container
+    # shellcheck disable=SC2086,SC2181
     rsync_output=$(kubectl exec $K8S_OBJECT -- rsync --version 2> /dev/null)
+    # shellcheck disable=SC2181
     if [ $? -eq 0 ]; then
         detected_rsync_version=$(echo "$rsync_output" | head -1 | awk '{ print $3 }')
-        if [[ "$detected_rsync_version" > 3 ]]; then
```
nicks, line 47: can you revert this? i think we did it deliberately this way because the alphabetical sort will work fine for what we want here

test.sh
```diff
@@ -3,5 +3,5 @@
 # find all scripts named 'test.sh' and run them
 # fail immediately if any fail
 
-cd "$(dirname "$0")"
+cd "$(dirname "$0")" || exit
```
nicks, line 6: ditto here

nicks (commented):
thank you for trying to wrangle this mess!

syncback/krsync.sh
dbast: done

Makefile
dbast: removed the complex command. will explain in another comment why/how.

Makefile
dbast: found a different solution. will explain in a further comment what/why.

.circleci/config.yml
dbast: done. ran into other issues (volume mounts work different than expected). Found different solution. Will explain that in a further comment.

dbast:
@nicks Went a different route: The command in the make target was complex and also didn't work as volume mounts behave different than expected on circle-ci.

Instead enabled to run shellcheck + other hooks via [pre-commit](https://pre-commit.com/):
* Extensible framework that installs, caches, executes various `hooks` to format, style-check, lint code
* All the run logic via e.g. `git ls-files` to apply only to repo of this repo, exclusion of files, reporting of an overall error code is thus handle by pre-commit
* pre-commit can despite its name be used without being a git pre-commit hook.
* It became super popular in recent years to aggregate linters via single .pre-commit-config.yaml files
* Enabled to run via circle ci with hook caching -> super fast execution ~22secs see https://circleci.com/gh/tilt-dev/tilt-extensions/4184
* There are lots of hooks that could be relevant for the repo here (can propose those via separate PRs): starlark formatter/linter, ruff to format / lint python code, shfmt to format shell scripts, hadolint to lint docker files, ymlformat to format yaml files... some of them are super nice, other maybe a bit overdone.
* The `make pre-commit` target is also easy executable locally during dev work.

A full-fledged config can be e.g. seen here https://github.com/conda/conda/blob/main/.pre-commit-config.yaml

PR #621, opened 2024-12-02
Fix helm_remote for OCI registries and add test

dbast:
Using OCI charts currently fails on master with:
```
Error in local: command "helm repo add prometheus oci://ghcr.io/prometheus-community/charts" failed.
error: exit status 1
```


nicks (approved):
lgtm!

PR #622, opened 2024-12-08
Enable .py linting with ruff via pre-commit hook and fix findings

dbast:
Runs ruff with the F,E,W (=flake8 Python base line linter) + I (isort).


nicks (approved):
neat! thank you! :pray:

PR #625, opened 2024-12-23
Lint Tiltfiles with ruff and fix findings

dbast:
Enable linting Tiltfiles via ruff and fix findings. As Tilfiles/Starlark is a subset of Python, we need to disable some Ruff rules that don't apply (done in the pyproject.toml).

This finds unused variables, ambiguous named variables (l is hard to distinguish from 1 or L depending what IDE font is used), mixing of tab/spaces used for indentation, etc.  

The commit of this PR can be potentially added to a `.git-blame-ignore-revs` after merge, see also https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/

Update: Tiltfiles are now identified by pre-commit via https://github.com/pre-commit/identify/pull/494


.pre-commit-config.yaml
```diff
@@ -15,7 +15,8 @@ repos:
     hooks:
       - id: ruff
         args: [--fix]
-        exclude: ^.*/Tiltfile$
+        files: ^.*Tiltfile|^.*\.py$
+        types_or: [text]
```
dbast, line 19: Can be updated to `types_or: [python, starlark]` (without further `files` definition) once this gets accepted https://github.com/pre-commit/identify/pull/494 and released.

dbast: done

PR #631, opened 2025-03-18
uibutton: accept env var as Dict[str, str]

### LINKED ISSUE #630: uibutton cmd_button: pass env vars as dict
Hello 👋 ,

Big user of Tilt and a fan of the ui_button extension ❤ .

One small issue though is that throughout Tilt functions, environment variables are generally passed as a dict, for example with `local` or `local_resource`.
This is not the case for cmd_button which asks for a list of strings with `"key=value"`.

A dict is more convenient to use and manipulate than a list of strings where we have to build "key=value".
Moreover, when environment vars needs to be reused between `local` and `cmd_button` they need to get transformed which adds a bit of boilerplate.

The cmd_button could easily support both List["key=value"] and Dict[str,str] env vars definitions by detecting if env is a dict and creating an associated list to feed to `v1alpha1.cmd`

I'll attach the corresponding PR, hopefully it can get accepted.
Thanks !

gfeun:
Closes #630

### DIFF
--- a/uibutton/README.md
+++ b/uibutton/README.md
@@ -39,6 +39,7 @@ cmd_button(name='foo',
            # If you need env var expansion *within the command itself*
            # you'll need to run it via a shell.
            argv=['/bin/sh', '-c', './reseed_database.sh --shard="$SHARD"'],
+           env={"HOST": "localhost", "PORT": "5432"},
    )
 ```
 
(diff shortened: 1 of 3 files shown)


nicks (approved):
lgtm!

nicks (approved):
lgtm!

PR #636, opened 2025-05-20
Add build_dependencies flag to helm_resource

brian-bk:
The `update_dependencies` flag will auto-bump versions in the `Chart.lock`. This isn't desirable in all situations. However, it's still useful to have the tilt managed `helm_resource` download (build) the dependencies in the first place.

Note: I'm not a tilt maintainer and unsure anything else that would need to be done here. I might suspect some test could be worth doing as well.


nicks:
(sorry for the delay, i thought i had merged this earlier)

PR #641, opened 2025-09-01
Add run_once extension for single execution in Tilt sessions

adamovittorio:
Introduce the `run_once` extension to allow functions to execute only once during a Tilt session, preventing resource redeclaration errors across multiple Tiltfiles. For example, when including a common dependency in a [Many Tiltfiles and Many Repos](https://docs.tilt.dev/multiple_repos.html) scenario.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ All extensions have been vetted and approved by the Tilt team.
 - [`pulumi`](/pulumi): Install Kubernetes resources with [Pulumi](https://www.pulumi.com/).
 - [`pypiserver`](/pypiserver): Run [pypiserver](https://pypi.org/project/pypiserver/) local container.
 - [`restart_process`](/restart_process): Wrap a `docker_build` or `custom_build` to restart the given entrypoint after a Live Update (replaces `restart_container()`)
+- [`run_once`](/run_once): Execute a function only once during a Tilt session.
 - [`secret`](/secret): Functions for creating secrets.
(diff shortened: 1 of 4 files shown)


nicks:
i don't think this extension is a good idea.

all the use cases you give in the README and test have subtle bugs and bad race conditions in them. if a user edits their tiltfile in a tilt session, the resource in the run_once will get dropped from the environment, and the user will get errors.

This doc explains the tilt environment model in more detail: https://docs.tilt.dev/controlloop.html

adamovittorio:
Alright then, so if I understand correctly this is more of a Tilt engine missing feature, where specific resources must be unique and can't be defined multiple times. Handling the state in the Tiltfile is considered bad practice and can lead to bugs and race conditions.

When multiple tiltfiles define the same helm repository and are loaded by other tiltfiles (in a many repos, many tiltfiles scenario), a name conflict arises. What is the suggested approach to safely include other tiltfiles where a chart repository may already be declared?

Tiltfile `a` declares a `helm_repository` named `x` and uses one of the charts there to spin up a deployment.
Tiltfile `b` declares the same `helm_repository x` to spin up another deployment and it also loads tiltfile `a`, which causes an error saying that the `helm_repository x` resource exists already.

nicks:
you can define a `base` Tiltfile that both tiltfile `a` and tiltfile `b` load it. the base file will only load once.

PR #643, opened 2025-09-09
✨ feat(files_glob): Add files_glob and watch_glob functions for pattern expansion

netbrain:
Updated the `README.md` to include an entry for the `files_glob` extension.

The `files_glob` extension now supports expanding glob-like patterns into real file paths at Tiltfile load time, enhancing dependency management in builds. It includes two functions: `files_glob` for path expansion and `watch_glob` for file change detection, ensuring Tilt reloads as files are added.

 New tests have been added in `files_glob/test/Tiltfile` to validate functionality, including deduplication of file paths from overlapping patterns. The helper now utilizes bash/find for practical pattern resolution.


nicks (commented):
i don't think we should accept this approach as a general solution, sorry.

the problem is that watch_glob reloads the tiltfile whenever any file in the glob changes. we don't want changes to a docker context to cause a tiltfile re-execution

PR #652, opened 2025-11-07
Feat(dotenv) support multi-line values

### LINKED ISSUE #651: dotenv extension does not support multiline values
Usually .env files are allowed to contain new lines if the value is surrounded by double quotes. The simple line splitting that dotenv does not support that case.

Minimal reproduction case:
`.env`
```
TEST="some
text"
```
Result is an error `Error in fail: Invalid format for dotenv file: ...`, while the file should be parsed correctly and put `TEST="some\ntext"` in the environment.

AKarbas:
Closes #651 

Removes support for lines starting with a `"` and ending with a `'` — and removes the corresponding test-case.

Also updated the Readme for the extension: only double-quotes are supported for multi-line strings.

### DIFF
--- a/dotenv/README.md
+++ b/dotenv/README.md
@@ -3,6 +3,7 @@
 Author: [ln](https://github.com/tachiniererin)
 
 dotenv reads `.env` or another specified file and loads the key value pairs into the environment.
+Supports multi-line strings, but only wrapped with double-quotes (`"`).
 
 ## Usage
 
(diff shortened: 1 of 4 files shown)


nicks:
thanks for calling out the missing test.sh file! i added it in https://github.com/tilt-dev/tilt-extensions/pull/654

AKarbas:
> ok, it looks like the test caught an infinite loop in your script. can you investigate? 🙏

Thanks — fixed! Also updated the PR description.

nicks (approved):
lgtm!

PR #682, opened 2026-07-20
fix timeout failing to call fg

arrrnas:
`timeout 15 fg` fails because `timeout` tries to run `fg` as an external command and not the shell built-in (and results in command not found), since `cancel/test/test.sh` doesn't `set -e` this failure wasn't visible. In practice `grep` just runs after `./trigger.sh` making it prone to race conditions, adding a loop to retry grep up to 30s


arrrnas:
looks like this depends on #683 as this fix triggers the test that fails without the session cookie

nicks (approved):
lgtm!

PR #683, opened 2026-07-20
fix request missing session token

arrrnas:
since https://github.com/tilt-dev/tilt/pull/6776 requests need to include a session token, test was making a request without one and getting `missing session token` error because of it. Adding a request to get the token and including it in the original request.


nicks (approved):
thanks!

PR #685, opened 2026-07-21
pack: use the configured Docker host in lifecycle containers

arrrnas:
The Pack CLI respects `DOCKER_HOST`, but without  `--docker-host inherit` its lifecycle containers default to mounting `/var/run/docker.sock`.  `--docker-host inherit` ensures the lifecycle containers use the same daemon as the Pack CLI. When `DOCKER_HOST` is unset, Pack retains its standard socket behavior.


nicks (approved):
lgtm!

